PR #422, opened 2015-01-07
- fix debug package unique key handling.

adrianschroeter:
This is fixing the missing debug packages in current factory builds.


schaefi:
looks good will merge

PR #423, opened 2015-01-12
mediacheck: give the user 5s time to actually see the result (also helps...

DimStar77:
... openQA to catch it) - Failure waits for user input already

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -30,6 +30,7 @@ export UTIMER_INFO=/dev/utimer
 export bootLoaderOK=0
 export enablePlymouth=1
 export IFS_ORIG=$IFS
+export MEDIACHECK_OK_TIMER=5
 
 #======================================
 # lookup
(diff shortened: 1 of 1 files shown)


schaefi:
yeah makes sense :)

being picky but could you fix the indention (4 spaces), thanks

DimStar77:
Sure - I missed that vi so nicely auto-indented but used a tab instead of spaces.. .corrected.

schaefi:
ok, just another little thing, we should at least mention that we are waiting for 5sec and that one can continue by pressing any key. Thus could you add another Echo statement like

```
Echo "Press any key to continue (waiting 5 sec...)"
```

schaefi:
and maybe put the time in a variable

DimStar77:
ok, time is in a variable, and I changed the read call to be
-s -n 1 -t <timeout)

-s => do not echo whatever input we get from the CLI if any (just looks neater)
-n 1 => actually do as we advertise: ANY key is sufficient (otherwise read waits for EOL)

schaefi:
great :) will merge it

PR #424, opened 2015-01-12
Disable yum repos from centos-releases

T1loc:
Add yum option disablerepo for disable online repos installed with the centos-release package.


schaefi:
looks good, thanks

PR #427, opened 2015-01-14
- fix dependencies of s390 requires

jordimassaguerpla:
* zipl was never an own package, it is part of s390-tools
* syslinux does not exist on SLES 12 s390

(cherry picked from commit 43798988bfd41072e6c537211b7360969fe1baf4)

### DIFF
--- a/rpm/kiwi.spec
+++ b/rpm/kiwi.spec
@@ -350,7 +350,7 @@ Requires:       yaboot
 %endif
 %endif
 %ifarch s390 s390x
-Requires:       zipl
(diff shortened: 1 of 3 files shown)


schaefi:
are you sure syslinux exists on s390 ?

jordimassaguerpla:
syslinux does not exist on s390. Actually that was an "unresolvable" on the build service when trying to build an s390x image.

schaefi:
ah yes sorry I misread the patch, though you have added it but you fixed it by removing it :)
all good thanks will merge it

PR #428, opened 2015-01-16
sha256sum generation: change cwd to tmpdir and only pass filename to $sh...

DimStar77:
...a->addfile

The full path passed to $sha->addfile is written into the .sha256 file, which results in
the file not 'really' usable for verification by users. Passing only the relative filename
works around this.


schaefi:
looks good to me, thanks

will merge and prepare a release for Factory today

PR #429, opened 2015-01-18
Timezone fix

magist3r:
When copy timezone file to /etc/timezone and then change timezone using YaST, time is set wrong. So we need to use symlink for this.


schaefi:
The patch looks good to me, but could you rebase your branch to be put on top of the current master
and provide only the relevant commits with regards to the change set ?

Thanks

magist3r:
done.

schaefi:
thanks

PR #430, opened 2015-01-21
Update SLE12 JeOS template for ppc

schaefi:
The description was missing firmware packages and did
not use grub2 as the bootloader which is the default
for sles12 on ppc


schaefi:
@k0da could you review the patch. All the other changes of the file Thorsten has posted were already in the master branch. Thus it would be nice if you can give it a short smoke test with the master branch kiwi as well. Thanks a ton :)

k0da:
I'll take a look

k0da:
LGTM

PR #431, opened 2015-01-21
Switch to using appstream-builder instead of XML parsing

DimStar77:
I would like to get a review on this work so far.

I managed to build the non-oss tree using this switch and get a good set of valid appstream metadata out of this.
The advantage over extracting .appdata during build is that this version can 'redo' the entire extraciton over all existing RPMs in case of issues, without having to rebuild the packages (think: extractor problem producing invalid data.. before: fix extractor, rebuild all packages, rebuild ftp tree; new: fix extractor (appstream-builder), rebuild ftp-tree

The variant for rpm-md is not yet working, as I did not (yet) prepare a test setup for it... If prefered, I can leave the existing extractor in place, only with a FIXME, and only switch the yast style repository creation.

### DIFF
--- a/distribution/kiwi-instsource-plugins-openSUSE-13-2/KIWIDescrPlugin.pm
+++ b/distribution/kiwi-instsource-plugins-openSUSE-13-2/KIWIDescrPlugin.pm
@@ -233,23 +233,18 @@ sub executeDir {
         );
         return 1;
     }
-    if ((-x "/usr/bin/extract-appdata-icons") &&
-        (-s "$targetdir/appdata.xml")
-    ) {
-        $cmd = "/usr/bin/extract-appdata-icons "
-            . "$targetdir/appdata.xml $targetdir";
-        $call = $this -> callCmd($cmd);
-        $status = $call->[0];
-        if($status) {
+    if (-x "/usr/bin/openSUSE-appstream-process") {
+        foreach my $p (@{$paths}) {
+            $cmd = "/usr/bin/openSUSE-appstream-process";
+            $cmd .= " $p";
+            $cmd .= " $targetdir";
+            $call = $this -> callCmd($cmd);
+            $status = $call->[0];
             my $out = join("\n",@{$call->[1]});
-            $this->logMsg("E",
+            $this->logMsg("I",
                 "Called <$cmd> exit status: <$status> output: $out"
             );
-            return 1;
-        }
-        if($this->{m_compress} =~ m{yes}i) {
-            system("gzip", "--rsyncable", "$targetdir/appdata.xml");
-        }
+        };
     }
     if($this->{m_compress} =~ m{yes}i) {
         foreach my $pfile(glob("$targetdir/packages*")) {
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks for your work, this touches the installation source code. Thus I'd like to ask Adrian for the review
@adrianschroeter Could you take a look ?

Thanks

DimStar77:
Finished the work to also work on repomd repositories - and tested on the non-oss tree as well (changed the kiwi file to create repomd information).

schaefi:
to me this looks all good, but Adrian has to give his ok as this code directly touches his work
I'll ping him

DimStar77:
Sure - thanks a lot (also for your first glance at it).

adrianschroeter:
Better ask @mlschroe, he has worked on that in the past....

mlschroe:
Well, one little problem I see is that I'm not really keen on calling appstream-builder from the OBS publisher. So there's a chance that the repositories generated by  the publisher contain very different appdata than the ones generated by kiwi.

(But you probably already knew that, as we had this discussion on IRC some time ago.)

I'm also not sure why we should hardcode all the calls in kiwi, appstream-builder could be called from the extract-appdata-icons switch, no? (Maybe we should rename that script to postprocess-appdata?)

DimStar77:
I am aware you don't want to start appstream-builder from the publisher; but the current scripts are broken beyond repair if you ask me (there are assumptions we just should not make, like extension packages) and catching up with the metadata modifications is just painful and useless, if we can use/share tools (my opinion)

The extract-appdata-icon script would no longer be needed on Factory builds at all.
changing the extract-appdata-icons to use appstream-builder is a no-go, as it would pull in the entire glib toolkig into each build (as extract-appdata-icon nis a Support package).

So putting the thing directly into kiwi makes sense from that PoV. What I 'could' do though would be  wrapper, part of the appstream-glib package, that takes care of the more special parameters (like api-version and so on), so that those can be updated/changed without touching kiwi.

DimStar77:
I added a 2nd commit on this branch which would be calling an extrenal script (already written in my OBS branch) which does the processing.

Th advantage would be that this can be updated without touching kiwi.

This entire change does not rely at all on the existing appstrean-extract perl scripts (also not changing them)

DimStar77:
@mlschroe is the extended version with the 2nd commit, changing the logic to live in an external script (thus easier to be updated) more to your liking?

schaefi:
@mlschroe any news ?

mlschroe:
Could you please merge the two commits?

(I also must admit I don't understand the old code, why is extract-appdata-icons called twice? Wouldn't it be saner to copy the appdata.xml into newtargetdir first if it doesn't exist there?)

Btw, what is $paths[0] and why wasn't it needed before?

As m_compress is a kiwi option, shouldn't /usr/bin/openSUSE-appstream-process leave the appdata.xml uncompressed?

DimStar77:
On Mon, 2015-02-09 at 05:55 -0800, Michael Schroeder wrote:
> Could you please merge the two commits?

Merged

> (I also must admit I don't understand the old code, why is
> extract-appdata-icons called twice? Wouldn't it be saner to copy the
> appdata.xml into newtargetdir first if it doesn't exist there?)
> 
> Btw, what is $paths[0] and why wasn't it needed before?

From my tests it was pointing to the directory with all the RPMs for the
repository to be created; exactly what is needed for appstream-builder
(which is what the openSUSE-appstream-process script calls)

> As m_compress is a kiwi option,
> shouldn't /usr/bin/openSUSE-appstream-process leave the appdata.xml
> uncompressed?

hmm. appstream-builder output an appdata.xml.gz and a app-icons.tar.gz;
there is no tool know that would need those files in uncompressed form.
of course I can decompress them after they are created, then recompress
them from within kiwi, but that sounds a bit backwards.


-- 
Dimstar / Dominique Leuenberger <dimstar@opensuse.org>

mlschroe:
Well, createrepo is called for every element in @paths, that would indicate that you also need to have multiple appstream-process calls. But for the "susetags" case, only one call to create_package_descr is done, creating only one repository in $paths[0]. The current code is a complete mess.

DimStar77:
On Mon, 2015-02-09 at 06:41 -0800, Michael Schroeder wrote:
> Well, createrepo is called for every element in @paths, that would
> indicate that you also need to have multiple appstream-process calls.
> But for the "susetags" case, only one call to create_package_descr is
> done, creating only one repository in $paths[0]. The current code is a
> complete mess.

in my tests over the Non-OSS repo, in both cases (yast-style repo and
reopmd) it did create what I needed, apparently because ExecuteDir is
already executed 'per directory'

-- 
Dimstar / Dominique Leuenberger <dimstar@opensuse.org>

mlschroe:
ExecuteDir is called per directory-set. In most cases the set seems to contain only one dir.

DimStar77:
On Mon, 2015-02-09 at 07:28 -0800, Michael Schroeder wrote:
> ExecuteDir is called per directory-set. In most cases the set seems to
> contain only one dir.

I see; I'll change the repomd case to iterate over the directories in
this case... for the unknown case that this might happen :)
-- 
Dimstar / Dominique Leuenberger <dimstar@opensuse.org>

DimStar77:
Is there anything else I can / should change?

schaefi:
to me this looks like a clean patch now.

@mlschroe can we merge it ?

schaefi:
I think we gain more attraction to this if we merge the code ;) the changes looks sane to me and simplify the code, let's check it in

mlschroe:
(sorry, I had a couple of days off)

PR #432, opened 2015-01-29
- collecting of -debuginfo packages from special suffixed packages

adrianschroeter:
This is fixing the qemu-x86 special case and simplifies the code.

It does also simplify the code, but it means also that this will not work on SLE 11 anymore. So please merge only to SLE 12 and master.

thanks!


schaefi:
ok, for sle12 I need another pull request. The sle12 branch is: KIWI-506-SuSE-13-1-SLE-SP-Devel

PR #435, opened 2015-02-03
aarch64: Load full EFI emulation size

dirkmueller:
-boot-load-size 1 is not a recommended option. First of all
a loader size that is not divideable by 4 is rejected by
most EFI firmwares. In addition we need to load all of the
EFI code into memory. When the option is not set, genisoimage
uses the full length by default, which is what we want.


schaefi:
looks good to me

just wondering why we have that value for x86_64_efi and no complains

PR #438, opened 2015-02-12
remove plymouth packages from s390x boot image

jordimassaguerpla:
We don't need graphical boot on s390x.


schaefi:
could you add the change to netboot / oemboot suse-SLES12 and suse-SLED12 as well ?

jordimassaguerpla:
done.
no SLED for s390x

schaefi:
great thanks

PR #439, opened 2015-02-13
check user_name is defined when checking file tree

jordimassaguerpla:
It is possible to have files that should be owned by root which
are owned by a non-existing user, meaning the uid has no corresponding
username. When this happens, the check used to give the error

"Use of uninitialized value $user_name in concatenation (.) or string
at /usr/share/kiwi/modules/KIWIRuntimeChecker.pm line 1558."

Thus we need to check whether the user exists.


schaefi:
This references bug bnc: 916509

just adding this for reference

schaefi:
lgtm

PR #446, opened 2015-03-04
- Fix consistency check for LXC builds

rjschwei:
+ Prior to the emergence of systemd the lxc package provided a special
    init script, lxc-init. This was necessary as SysV-init was not container
    aware. systemd is container aware, thus including systemd as the init
    system is sufficient.


schaefi:
Thanks, looks good to me

PR #447, opened 2015-03-05
dhclient has no option for timeout

cschneemann:
The dhclient in openSUSE 13.2 has no parameter for timeout, but it
is specified in the setupNetworkDHCLIENT function. The timeout is
taken from the configuration file as mentioned in the manpage.


schaefi:
Thanks, good catch

PR #448, opened 2015-03-17
Remove KIWI's appdata handling

DimStar77:
The format is just too complex to be handled in this symlistic way.
Appdata.xml (a subset of Appstream) evolved and changed for example
from "applications" to "components"; now, some packages export one type,
other packages export the other type. Merging them without normalising
is resulting in a structure like:
```
<applications>
  <application>
    [...]
  </application>
  <component>
    [...]
  </components>
</applications>
```
whereas 'components' is no valid child of applications per spec (neither
is application a valid child of components).

Hence: if KIWI ought to do that proper, the full format must be
understood, analysed, normalised and merged.

for openSUSE, KIWI now uses appstream-builder, which is created in close
relationship with the AppStream metadata in mind.

Having this code still in KIWI results in the builder producing valid
appdata.xml.gz PLUS an uncompressed appdata.xml (~18MB) to be part of
the repository metadata (thus slowing repo refresh down a lot).

### DIFF
--- a/modules/KIWICollect.pm
+++ b/modules/KIWICollect.pm
@@ -118,10 +118,6 @@ sub new {
         m_fdebugpacks  => [],
         m_debug        => undef,
         m_rmlists      => undef,
-        m_appdata      => undef,
-        m_appdata_seen => undef,
-        m_appdata_base => undef,
-        m_appdata_type => undef,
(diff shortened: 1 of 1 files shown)


schaefi:
This sounds very good to me

PR #449, opened 2015-03-18
Changes to instsource requirements

k0da:
In order to make syslinux requirement conditional we need to build
kiwi-instsource arch specific package.

This commit is needed because we can't Require kiwi-isoboot in Factory
anymore since we started to build PowerPC in Rings


schaefi:
looks good, will merge and submit to master in a minute

PR #450, opened 2015-03-23
Complete KIWI's appdata handlling removal

DimStar77:
Commit f4b7df2 removed the general handling, but missed a reference in
KIWICollect.pm

Sorry for that one


schaefi:
ok :)

PR #452, opened 2015-04-10
- adapt sort order for KIWICollect:

adrianschroeter:
1. prefer repository by priority
 2. sort by architectures and their fallbacks

for bnc#926651


schaefi:
the indention seems broken at several places, could you fix this by the correct amount of spaces ? Thanks

schaefi:
looks good now, thanks

PR #456, opened 2015-04-22
Added __checkMountDependencies runtime check

schaefi:
The check looks for processes which holds onto mounts created by a previos kiwi run. In such a condition kiwi is not able to release its own mount table and shows information about the processes which keeps these mount points busy. We still don't understand why there could be non kiwi related processes holding onto mounts kiwi has created during its run but kiwi should provide a way to tell this information to the user and stop if the environment is still tainted by such processes

https://bugzilla.suse.com/show_bug.cgi?id=895204

### DIFF
--- a/modules/KIWIRuntimeChecker.pm
+++ b/modules/KIWIRuntimeChecker.pm
@@ -1486,30 +1486,56 @@ sub __checkMountDependencies {
     if ($ENV{KIWI_IGNORE_OLD_MOUNTS}) {
         return 1;
     }
-    my $kiwi_mount_references = KIWIQX::qxx (
-        "grep -il kiwi /proc/[1-9]*/task/*/mounts|cut -d/ -f3|sort -u 2>&1"
-    );
-    if (! $kiwi_mount_references) {
+    my %proc_result;
+    foreach my $pid ($this->__read_pids('/proc')) {
+        foreach my $task_pid ($this->__read_pids("/proc/$pid/task")) {
+            my $TASK_MOUNTS = FileHandle->new();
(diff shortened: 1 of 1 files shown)


modules/KIWIRuntimeChecker.pm
```diff
@@ -1470,6 +1477,43 @@ sub __checkDeprecatedFilesystem {
...
+#------------------------------------------
+sub __checkMountDependencies {
+    my $this = shift;
+    my $kiwi = $this->{kiwi};
+    my $msg;
+    if ($ENV{KIWI_IGNORE_OLD_MOUNTS}) {
+        return 1;
+    }
+    my $kiwi_mount_references = KIWIQX::qxx (
+        "grep -il kiwi /proc/[1-9]*/task/*/mounts|cut -d/ -f3|sort -u 2>&1"
```
aspiers: I wasn't sure if `kiwi` was the best thing to `grep` for here - but you know better than me :)

aspiers:
LGTM.  Personally I would much prefer a heredoc for the large chunk of text, but that's just a stylistic choice ...

schaefi:
yep let's make this a little bit nicer by using here messages and prevent invoking too much shell code :)

schaefi:
how about it now ?

by the way I use the output from pstree to show the processes in question, imho that gives a nicer overview even though the tree indentation might become broken because only parts of it are shown, a cosmetic thing I could live with

schaefi:
with regards to the search expression you are right. There is no 100% safe way to match a kiwi process, but looking up 'kiwi' in the name is fairly good because I use that as prefix for every mount in the code. Of course any other unrelated process could hold on a mount with kiwi in its name. They would be false positives

PR #459, opened 2015-04-29
ARM efi is different

rjschwei:


schaefi:
I'm sorry but this is not going to work reliably. The patch checks the host architecture and will come to the wrong decision for a cross platform build. This needs to be implemented as grub script checking for 

```
if [ "${grub_cpu}" = "x86_64" ] || [ "${grub_cpu}" = "i386" ]; then
```

rjschwei:
Fair enough, fixed. Although in the past we have always proclaimed that cross platform image builds were not a goal. Considering that we have difficulties (not necessarily due to kiwi design and implementation) getting builds to work consistently if the target image distro and the build system distro are different I am not convinced cross platform image builds will be any easier to get working.

schaefi:
yes but for arm and also aarch64 people spent some time in host emulation just because many of those devices are not performant enough for native builds. So in this environment building arm images on x86_64 was a common use case. If implemented correctly uname would also report the right emulated architecture information but I don't want to rely on that.

Especially with regards to the boot loader I find it more stable to let the bootloader provide me the architecture information.

Thanks for adapting the patch, it should also be less code now :)

schaefi:
ok looks all good

PR #460, opened 2015-05-05
Fix lvm volume regex to allow volume names that are just have a forbidde...

bgeuken:
...n name as prefix


schaefi:
Test will follow, no merge until then

PR #461, opened 2015-05-12
Google doesn't like pretty print of json metadata

schaefi:


schaefi:
with this commit the license information looks like this

```
{"licenses":["0815"]}
```

PR #462, opened 2015-05-12
Introduce initial support for App Container Image (aci)

matwey:
Here is initial implementation of App Container Image (aci) format.
This commit is able to build .aci images conforming to SPEC.

[1] https://github.com/appc/spec/blob/master/SPEC.md#app-container-image

This commit writes all required fields to manifest, but not support (yet) optional fields, like app. So, there is not way (yet) to specify which application to start. However, further requests should follow.

### DIFF
--- a/modules/KIWIACIConfigWriter.pm
+++ b/modules/KIWIACIConfigWriter.pm
@@ -1,0 +1,111 @@
+#================
+# FILE          : KIWIACIConfigWriter.pm
+#----------------
+# PROJECT       : openSUSE Build-Service
+# COPYRIGHT     : (c) 2015 Matwey V. Kornilov
+#               :
+# AUTHOR        : Matwey V. Kornilov <matwey.kornilov@gmail.com>
+#               :
+# BELONGS TO    : Operating System images
+#               :
+# DESCRIPTION   : Write a manifest file for a ACI
+#               :
+# STATUS        : Development
+#----------------
+package KIWIACIConfigWriter;
+#==========================================
+# Modules
+#------------------------------------------
+use strict;
+use warnings;
+use JSON;
+
+#==========================================
+# Base class
+#------------------------------------------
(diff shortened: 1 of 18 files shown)


modules/KIWIACIConfigWriter.pm
```diff
@@ -0,0 +1,111 @@
...
+    "name" => $name,
+    "labels" =>
+    [{
+         "name" => "os",
+         "value" => "linux",
+    },
+    {
+         "name" => "version",
+         "value" => $iver,
+    }],
```
rjschwei: Please use single quotes, i.e. ' for all strings that are not interpolated. I know kiwi is not yet consistent in that way but we are trying to get there.

rjschwei:
- New files need to be added to: .git-hooks/pre-commit
- You should also create the following link .git/hooks -> ../.git-hooks
- All files should pass at least perl-critic level 3
- There should be a test for the config write, see tests/unit/KTContainerConfigWriter.t and related files as an example

This is a great start

schaefi:
great start, thanks

please move 

```
git mv doc/examples/suse-13.1/suse-aci-container doc/examples/extras/suse-13.1
```

The reason is that the standard doc/examples follow the deprecation policy and thus will be
removed once 13.1 is no longer supported. Example templates like this should not be deleted even if the distro they refer to is no longer in the support matrix

schaefi:
you can get rid of the zypper setup in doc/examples/suse-13.1/suse-aci-container/root/etc/zypp/repos.d/ if you set the attribute imageinclude="true" to all the repo definitions in the kiwi config.xml. kiwi will then add the repos in the image and you can remove them from your overlay tree

matwey:
So, I am waiting for next comments.

schaefi:
This looks great to me now, I'm going to merge it

PR #476, opened 2015-08-03
SatSolver debugging

rjschwei:
Have the code ready for solver debugging.

### DIFF
--- a/modules/KIWISatSolver.pm
+++ b/modules/KIWISatSolver.pm
@@ -231,6 +231,8 @@ sub new {
         # Create SaT pool
         #------------------------------------------
         $pool = $this -> __createPool();
+        # Help with debugging solver issues
+        # pool->set_debuglevel(5);
         #==========================================
         # Create SaT repos
         #------------------------------------------


schaefi:
Is this still needed with regards to the information you got from Stefan and Michael ?

rjschwei:
On 08/06/2015 11:12 AM, Marcus Schäfer wrote:
> Is this still needed with regards to the information you got from Stefan and Michael ?
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/openSUSE/kiwi/pull/476#issuecomment-128406912
> 

Yes, I think it would be nice to have this comment in the code. Then it
is easy to generate solver debug output by simply uncommenting a line of
code in KIWI rather than having to know the syntax all the time ;)

Thanks,

-- 
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com
IRC: robjo

schaefi:
ok

PR #479, opened 2015-08-12
Remove genisoimage and cdrkit-cdrtools-compat dependency

jotrk:
* cdrkit-cdrtools-compat provides a symlink from mkisofs to genisoimage
* It should be safe to use mkisofs as drop-in replacement for genisoimage
* Removing these dependencies is one step towards removing wodim


schaefi:
The patch looks good to me, however I need to ask Steffen if this is save too

@wfeldt could you please review this change ?

Thanks

wfeldt:
the patch is ok

schaefi:
Thanks

PR #481, opened 2015-08-25
Revert "Remove genisoimage and cdrkit-cdrtools-compat dependency"

schaefi:
Reverts openSUSE/kiwi#479

### DIFF
--- a/doc/devdoc/devDocs/KIWIIsoLinux.html
+++ b/doc/devdoc/devDocs/KIWIIsoLinux.html
@@ -193,7 +193,7 @@ Called in
 <a name="new"><h2>new</h2></a>
 <p>
      Create a new KIWIIsoLinux object which is used to wrap
-     around the major mkisofs call. This code requires a
+     around the major genisoimage/mkisofs call. This code requires a
      specific source directory structure which is:
 <p>
 Defined on line: 46
@@ -215,7 +215,7 @@ Called in
 <li><a href="KIWIIsoLinux.html">KIWIIsoLinux</a> : 1174</li><li><a href="KIWIIsoLinux.html">KIWIIsoLinux</a> : 1357</li></ul>
 <a name="relocateCatalog"><h2>relocateCatalog</h2></a>
 <p>
-     mkisofs leave one sector empty (or fill it with
+     mkisofs/genisoimage leave one sector empty (or fill it with
      version info if the ISODEBUG environment variable is set) before
      starting the path table. We use this space to move the boot
      catalog there. It's important that the boot catalog is at the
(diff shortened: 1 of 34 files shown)


schaefi:
I'm sorry but the integration tests fails with the solution provided in #479 

zypper said:

```
Problem: cdrkit-cdrtools-compat-1.1.11-20.1.7.x86_64 conflicts with mkisofs provided by mkisofs-3.01~a24-2.2.1.x86_64
 Solution 1: do not install mkisofs-3.01~a24-2.2.1.x86_64
 Solution 2: do not install cdrkit-cdrtools-compat-1.1.11-20.1.7.x86_64

```

It's not a good idea to specify both

I'm sorry, reverting this

jotrk:
Hmmm.. I thought I removed all traces to cdrkit. Where does this dependency conflict come from?

All I can find is this:

```
# grep -r cdrkit *
system/boot/ix86/isoboot/suse-13.2/config.xml:        <package name="cdrkit-cdrtools-compat"/>
```

Which I'll remove, but I don't think that it solves the dependency issue.

schaefi:
It's from system/boot/ix86/isoboot/suse-13.2/config.xml, the config contained mkisofs and cdrkit-cdrtools-compat which caused the conflict. Only one is allowed

schaefi:
cdrkit-cdrtools-compat provides mkisofs, thus mkisofs should not be listed. In kiwi the first tool for iso creation is genisoimage and for backward compatibility we also can use mkisofs. The reason why the boot images contains cdrkit-cdrtools-compat and genisoimage is that it provides the isoinfo tool which is used by kiwi inside of the initrd. the compat package correctly links to the right isoinfo tool.

Adding mkisofs for SUSE should not be required at all, since sle11 is the oldest distro supported by kiwi

All in all I'm a little bit confused what we want to achieve ?

jotrk:
`cdrkit-cdrtools-compat` is based on `wodim`. `wodim` is not maintained since more than four or five years. I'd like to get rid of `wodim` in Factory, but cannot because of this dependency. `mkisofs` can also be found in `cdrtools`, `cdrkit-cdrtools-compat` actually only provides a symlink from `genisoimage` to `mkisofs`. Both versions are similar (though not completely), but `mkisofs` from `cdrtools` is actively maintained.
However, I don't know if there is an alternative to the `isoinfo` tool.

schaefi:
If we remove cdrkit-cdrtools-compat from the image descriptions kiwi should still be able to find isoinfo from /usr/lib/genisoimage/isoinfo. The kiwi code searches in both places. Thus you only have to make sure isoinfo is provided in either /usr/bin/ or /usr/lib/genisoimage

If we can make that sure cdrkit-cdrtools-compat could be deleted from the kiwi descriptions and also from the spec file.

However a compat package seems to be still required unrelated to kiwi, because of the symlinks to well known tools, e.g /usr/bin/mkisofs -> genisoimage

jotrk:
I'm not entirely sure about the compat package. Currently it only provides symlinks from {cd,dvd}record -> wodim etc. and of course from mkisofs -> genisoimage.
However, the point is, that users should use the {cd,dvd}record and mkisofs tools from cdrtools.
What I can imagine would be a cdrtools-cdrkit-compat package, providing symlinks in the other direction (wodim -> cdrecord, genisoimage -> mkisofs, etc.)

PR #482, opened 2015-08-31
Increase fsoverhead for btrfs again

andreas-schwab:


schaefi:
not sure if we are on the right track with this. However I have no problem to merge that because I think people using btrfs are not really interested to keep it small :)

I'm going to merge it as soon as I have finished my s390 work

Thanks

PR #483, opened 2015-09-01
Fix kernel name in extractLinux on s390x arch

mjura:
I have problem with extracting kernel on s390x architecture. As I have found problem is that on s390x link to kernel file is called `/boot/image`. 

This patch fix this issue

### DIFF
--- a/modules/KIWIImage.pm
+++ b/modules/KIWIImage.pm
@@ -4129,6 +4129,12 @@ sub extractLinux {
     if (-f "$imageTree/boot/vmlinux") {
         $kernname = "vmlinux";
     }
+    #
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks for your time and effort

the patch works however it should not be needed. if you look at
https://github.com/openSUSE/kiwi/blob/master/modules/KIWIConfig.sh#L1575

you can see that the image kernel will be extracted and is available as common name vmlinuz in the scope of kiwi.

So this kernel extraction is already taken care for

The problem you stumbled over seems to have another cause. I'll come back to your e-mail and ssh into the machine you mentioned

I'm sorry for the moment I can't merge this

PR #488, opened 2015-09-12
openSUSE 13.3 (current Tumbleweed suse_version)

DimStar77:
This should presumably fix the issues currently seen in openSUSE:Factory (see https://bugzilla.opensuse.org/show_bug.cgi?id=945584 )

the four suse-13.3 directories are copies of suse-13.2, only the config.xml files have been updated to reflect the new name.

Not sure if this is sufficient of if anything else is needed.


schaefi:
Thanks for looking into that however I have several concerns with our non enterprise strategy
especially with regards to image templates.

I'm pretty much confused about SUSE's non enterprise distributions. There is Factory, There is Tumbleweed considered to be kind of a stable factory and there is now also Leap42 whatever that means.

From my perspective it can't be even more complex to users given that there was a simple openSUSE distro out and stable for some time

In kiwi I adhere to the concept to provide templates for stable distributions with a specific live time in order to be able to provide a deprecation mechanism. That means for distros which do not change the kernel, it's structure or anything else in the system just because people submit changes to it. A template for such dynamic distros without a stable concept is basically always broken.

This is the reason why I'm still waiting for a clear picture about all this, and I must say I don't see one

Looking at your commit, I see you named the boot images for tumbleweed suse-13.3, however such a distributions does not exist. If tumbleweed will be the new openSUSE the boot image should be named suse-tumbleweed.

As you can see from the name there is now no version and obsolete policy possible. tumbleweed is a moving target, tumbleweed today is not tumbleweed from tomorrry

To be honest I don't want to maintain a description for a distro I have no idea how it looks like next week

Maybe I just misunderstood the concepts about our new open distro investment. In that case I just want to understand where the stable release is, how long it is considered to be stable and how I can apply a deprecation policy

Thanks

DimStar77:
>I'm pretty much confused about SUSE's non enterprise distributions. There is Factory, There is Tumbleweed considered to be kind of a stable factory and there is now also Leap42 whatever that means.

No, there is is only Tumbleweed and Leap - Tumbleweed is the rolling distribution, Leap is the stable release, based on SLE sources. Factory is not a distribution that exists; it's only the project name in OBS for the integration work.

> From my perspective it can't be even more complex to users given that there was a simple openSUSE distro out and stable for some time

We now just have two openSUSE products: a rolling one, daily up-to-date, and a stable one. The difference to earlier is that the stable one is not 'just' a random snapshot of the rolling release anymore (which, in fact it was... just that we had too much breakage in the rolling release (then called Factory) to advertise it for usage).

> From my perspective it can't be even more complex to users given that there was a simple openSUSE distro out and stable for some time

I went with suse-13.3 as suse_version in TW is 1330, which by all macros just translates to 13.3 in the kiwi config... so it was easiest to go with it.

Leap uses version 1315 (same as SLE12), which actually translates to suse-13.1 in Kiwi configurations, as far as I understood the logics. suse-tumbleweed sounds like a sensible name, but it means sue_Version 1330 needs to be specifically intercepted to adjust the naming pattern.

> Maybe I just misunderstood the concepts about our new open distro investment. In that case I just want to understand where the stable release is, how long it is considered to be stable and how I can apply a deprecation policy

Leap 42.1 will be the next stable release, with SLE12SP12 as it's base, overloaded with a bunch of packages / apps / desktops from Tumbleweed... so we get stable as SLE, new as Tumbleweed (sort of). The lifetime of 42.1 is in alignment with SLE12SP1; 42.2 will be released with SLE12SP2 as base.

schaefi:
Talking to some people I think Leap42 is the new openSUSE

coolo:
a moving suse-tumbleweed config sounds reasonable to me

schaefi:
ok I now understand why you chose 13.3 as the name. However this could be addressed in the spec file which builds the internal -requires packages used by the buildservice. Looking there I see the
following definition:

%if %suse_version == 1315
    %define mysystems suse-SLES12

So for tumbleweed it would be easy to match that against suse-tumbleweed, however how can I distinguish Leap from SLES ? or can it be guaranteed that Leap and SLES will be compatible in terms of the kernel and kernel structure (this is the most critical part for me at this point)

Thanks

DimStar77:
SLE12:
%suse_version == 1315 && !0%{?is_opensuse}

Leap 42.1:
%{suse_version} == 1315 && 0%{?is_opensuse}

schaefi:
thanks

DimStar77:
Leap 42.1 currently uses kernel 4.1; so it's definitively not the same as SLE there.

schaefi:
yes that's what I thought too, thanks for double checking

schaefi:
ok I'm going to adapt the changes in the spec file and will also add a leap42 template.

Could you change your commit to use suse-tumbleweed as boot path ?

We also should add a tumbleweed example template in templates/ and add to the description that this is a moving target and what you build today is not necessarily the same as what you build tomorrow. I would also add the information that we are not providing any warranty for the tumbleweed build as long as we don't have an integration test matrix building with the same speed like tumbleweed moves

Thoughts ?

schaefi:
ok, done, you should be ready to update your commit. Thanks much

schaefi:
Thanks, looks good to me

PR #489, opened 2015-09-13
Add oemboot support for SLE12/aarch64 (fate#318444)

dirkmueller:


schaefi:
Thanks much, looks good to me

PR #491, opened 2015-09-17
Add vmxboot support for aarch64 on SLE12 and leap42.1

dirkmueller:


schaefi:
Hmm, for what reason do we need vmxboot on arm, actually we only builds oem's

dirkmueller:
@schaefi for building a KVM JeOS on aarch64 (see the JeOS efforts happening on SLE12 base right now)

schaefi:
ok because of the JeOS product

PR #492, opened 2015-09-18
- Add "toram" boot parameter to allow running the live system entirel…

cyberorg:
…y from RAM

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -6971,15 +6971,15 @@ function mountSystem {
 #--------------------------------------
 function mountOrCopyLiveCD {
     if [ -n $TORAM ]; then
-	Echo "Copying CD system into tmpfs"
+        Echo "Copying CD system into tmpfs"
         mkdir -p ${LIVECD}R $LIVECD && eval mount $cdopt $biosBootDevice ${LIVECD}R
         SIZE="$(du -s /${LIVECD}R | gawk '{print int($1*1.1)}')"
         mount -t tmpfs -o size=${SIZE}k tmpfs $LIVECD
         cp -ar ${LIVECD}R/* $LIVECD/
         umount ${LIVECD}R
-	rmdir ${LIVECD}R
+        rmdir ${LIVECD}R
     else
-	mkdir -p $LIVECD && eval mount $cdopt $biosBootDevice $LIVECD
+        mkdir -p $LIVECD && eval mount $cdopt $biosBootDevice $LIVECD
     fi
 }
 #======================================
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -6971,11 +6971,13 @@ function mountSystem {
 #--------------------------------------
 function mountOrCopyLiveCD {
     if [ -n $TORAM ]; then
+	Echo "Copying CD system into tmpfs"
         mkdir -p ${LIVECD}R $LIVECD && eval mount $cdopt $biosBootDevice ${LIVECD}R
-        SIZE="$(/usr/bin/du -s /${LIVECD}R | /usr/bin/gawk '{print int($1*1.1)}')"
-        /bin/mount -t tmpfs -o size=${SIZE}k tmpfs $LIVECD
+        SIZE="$(du -s /${LIVECD}R | gawk '{print int($1*1.1)}')"
+        mount -t tmpfs -o size=${SIZE}k tmpfs $LIVECD
         cp -ar ${LIVECD}R/* $LIVECD/
         umount ${LIVECD}R
+	rmdir ${LIVECD}R
     else
 	mkdir -p $LIVECD && eval mount $cdopt $biosBootDevice $LIVECD
     fi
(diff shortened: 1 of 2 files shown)


modules/KIWILinuxRC.sh
```diff
@@ -6967,6 +6967,21 @@ function mountSystem {
     return $retval
 }
 #======================================
+# mountLivecd
+#--------------------------------------
+function mountLivecd {
+    if [ -n $TORAM ]; then
+        mkdir -p ${LIVECD}R $LIVECD && eval mount $cdopt $biosBootDevice ${LIVECD}R
+        SIZE="$(/usr/bin/du -s /${LIVECD}R | /usr/bin/gawk '{print int($1*1.1)}')"
+        test -n $SIZE || SIZE="800000"
```
schaefi: I think it can't happen that SIZE is empty and if it happens we should log this and proceed with the normal way of mounting the CD

modules/KIWILinuxRC.sh
```diff
@@ -6967,6 +6967,21 @@ function mountSystem {
...
+        /bin/mount -t tmpfs -o size=${SIZE}k tmpfs $LIVECD
+        cp -r ${LIVECD}R/* $LIVECD/
```
schaefi: Instead of cp please use rsync, that avoids problems with file permissions. rsync is available in the initrd

system/boot/ix86/isoboot/suse-linuxrc
```diff
@@ -65,6 +65,9 @@ includeKernelParameters
 if grep -i -q "livecd_config=[a-zA-Z0-9/]" /proc/cmdline; then
     export LIVECD_CONFIG=`sed 's/.*livecd_config=\([^ ]*\).*/\1/' /proc/cmdline`
 fi
+if grep -qi "toram" /proc/cmdline; then
+        export TORAM=true
```
schaefi: you export TORAM but I don't see where this variable is used ?

modules/KIWILinuxRC.sh
cyberorg: you can remove that line, that was copy/paste leftover from knoppix linuxrc

system/boot/ix86/isoboot/suse-linuxrc
```diff
@@ -210,7 +213,7 @@ if [ -z "$UNIONFS_CONFIG" ] && [ -z "$COMBINED_IMAGE" ];then
         systemException "Failed to mount RW root filesystem" "reboot"
     fi
     cdopt=$(CDMountOption $biosBootDevice)
-    mkdir -p $LIVECD && eval mount $cdopt $biosBootDevice $LIVECD
+    mountLivecd
```
schaefi: the name is misleading, maybe better mountOrCopyLiveCD or embedLiveCD or integrateLiveCD

schaefi:
Please also add a short information about this feature in the documentation. In the branch

```
improve_documentation
```

Thanks

modules/KIWILinuxRC.sh
cyberorg: will use cp -ar as rsync would have to be added to boot image otherwise. We copy ro image anyway so permissions are not critical.

system/boot/ix86/isoboot/suse-linuxrc
cyberorg: in KIWILinuxRC.sh function

system/boot/ix86/isoboot/suse-linuxrc
cyberorg: mountOrCopyLiveCD sounds more descriptive

cyberorg:
Please add this in the docs, I am not sure about place it goes in the git branch, it can come under kiwidebug parameter or somewhere under isoboot/live CD section.

The boot option toram copies contents of live CD into RAM to allow running the live system from RAM 
entirely. CD or USB stick can be ejected after boot completes.

modules/KIWILinuxRC.sh
```diff
@@ -6967,6 +6967,20 @@ function mountSystem {
...
 #======================================
+# mountOrCopyLiveCD
+#--------------------------------------
+function mountOrCopyLiveCD {
+    if [ -n $TORAM ]; then
+        mkdir -p ${LIVECD}R $LIVECD && eval mount $cdopt $biosBootDevice ${LIVECD}R
+        SIZE="$(/usr/bin/du -s /${LIVECD}R | /usr/bin/gawk '{print int($1*1.1)}')"
+        /bin/mount -t tmpfs -o size=${SIZE}k tmpfs $LIVECD
+        cp -ar ${LIVECD}R/* $LIVECD/
+        umount ${LIVECD}R
```
schaefi: looks good, just add 

```
rmdir ${LIVECD}R
```

and it would be nice to tell the user what we are doing here. So a message saying

```
Echo "Copying CD system into tmpfs"
```

or similar

Thanks

schaefi:
I have created a pull request against the improve_documentation branch and hope Frank will merge it

schaefi:
all right, let's go with it :) Thanks

PR #495, opened 2015-09-24
Added arm suse-tumbleweed vmx|oemboot descriptions

schaefi:
Tumbleweed boot descriptions for arm


schaefi:
@dirkmueller hi could you have a look at it, the oemboot data is the one from Andreas

Thanks

schaefi:
I guess it's ok to merge, at least I could not find a problem in the file lists so I'd like to go ahead and submit a new kiwi version to factory

PR #496, opened 2015-09-24
make vmx image testable in openQA

lnussel:


schaefi:
looks good thanks much

PR #497, opened 2015-09-24
Fix grub2-mkimage on PowerPC

k0da:


schaefi:
Wohoo, great fix :) thanks much

PR #498, opened 2015-09-25
Set PReP device if not defined

k0da:
This fixes vmxboot, where ImagePrepDevice is not defined

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -1324,8 +1324,9 @@ EOF
     # install grub2 in BIOS mode
     #--------------------------------------
     if [ ! -z "$kiwi_OfwGrub" ];then
+        local prepdev=$(ddn $imageDiskDevice $kiwi_OfwGrub)
         # install powerpc grub2
-        $instTool $imagePrepDevice 1>&2
+        $instTool $prepdev 1>&2
         if [ ! $? = 0 ];then
             Echo "Failed to install boot loader"
             return 1
--- a/system/boot/ppc/oemboot/suse-repart
+++ b/system/boot/ppc/oemboot/suse-repart
@@ -766,7 +766,6 @@ function setupDeviceNames {
     local bootID=$4
     local iorwID=$5
     local vgroup=$6
-    local prepID=$7
     if [ -z "$vgroup" ];then
         #======================================
         # set root device name
@@ -801,10 +800,6 @@ function setupDeviceNames {
         # set boot device name
         #--------------------------------------
         export imageBootDevice=$(ddn $imageDiskDevice $bootID)
-        #======================================
-        # set PReP device name
-        #--------------------------------------
-        export imagePrepDevice=$(ddn $imageDiskDevice $prepID)
         #======================================
         # set boot partition id
         #--------------------------------------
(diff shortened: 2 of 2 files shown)


modules/KIWILinuxRC.sh
```diff
@@ -1324,8 +1324,10 @@ EOF
     # install grub2 in BIOS mode
     #--------------------------------------
     if [ ! -z "$kiwi_OfwGrub" ];then
+        local prepdev=$imagePrepDevice
+        test -z $prepdev && prepdev=$(ddn $imageDiskDevice $kiwi_OfwGrub)
```
schaefi: I realized that imagePrepDevice is never set by kiwi. This means we either lost something in the past or that code path never worked.

Can you confirm that we need the prep device only for the purpose of installing grub correctly ? if so you can delete the use of $imagePrepDevice entirely and just write

```bash
prepdev=$(ddn $imageDiskDevice $kiwi_OfwGrub)
```

k0da: It is used by oemboot.

schaefi: a grep search on the master code did not match anything

k0da: system/boot/ppc/oemboot/suse-repart:        export imagePrepDevice=$(ddn $imageDiskDevice $prepID)
system/boot/ppc/oemboot/suse-repart:        export imagePrepDevice=$(ddn $imageDiskDevice $prepID)

k0da: https://github.com/openSUSE/kiwi/blob/master/system/boot/ppc/oemboot/suse-repart#L848

schaefi: ah, my fault sorry I searched in the wrong architecture branch

you are right of course

schaefi: but this code in kiwi is still broken, the ppc oemboot suse-repart has a function named

```
setupDeviceNames rootID swapID recoID bootID iorwID vgroup prepID
```

The prep id which causes imagePrepDevice to become exported is the seventh parameter.
However this parameter is never set in any of the function calls

Thus I would expect this variable to be empty all the time. actually this code should be cleaned up. imagePrepDevice is only exported in suse-preinit but never used. The only use of the variable is the code part you have changed.

Given that the variable is not useful at all I suggest we delete it from suse-preinit and in KIWILinuxRC.sh and only go with validating $kiwi_OfwGrub

Thoughts ?

k0da: Well, it is used in KIWILinuxRC.sh

the code was added by me and I'm sure I've tested it.

schaefi: I believe you that your code works, but why do we want to mess with a variable which is not needed at all. Your code adds

```
test -z $prepdev && prepdev=$(ddn $imageDiskDevice $kiwi_OfwGrub)
```

but test -z $prepdev is always true because suse-preinit never exports imagePrepDevice
Thus why not simply delete the imagePrepDevice mess

This is not about your code it's about my crappy code :)

k0da: @schaefi Looking at commit, it seems you're rigth.. I'll redo the patch

k0da:
updated

schaefi:
Thanks this looks good to me, have you tested that part ?

PR #500, opened 2015-10-01
/run should not be mounted in firstboot initrd for distros not using systemd

mmohring:
CentOS, RHEL 6.0' or similiar should be handled like SLE 11 not using a tmpfs on /run in the initial initrd created by KIWI.


mmohring:
Pre systemd distributions like RHEL 6, CentOS 6 and SLE 11 should not mount /run and /var/run to a tmpfs. It makes the first boot fail because /var/run is then not correctly setup.

schaefi:
Thanks much for catching this one, looks good to me

PR #502, opened 2015-10-02
Add REPO_LOCATION to grub.cfg

k0da:
Patch grub2.cfg based mini isos to insert repo location provided
by _product


k0da:
Ah it is already there:
https://github.com/openSUSE/kiwi/pull/441

PR #508, opened 2015-10-28
Added vmdk descriptor update

schaefi:
VMWare applications are made aware that an image has vm-tools installed
via tags appended to the VMDK's metadata. Typically, the tools write
the metadata during their first run, but in our case the first run of
the tools is the first boot of the appliance, so the host application
is not aware of tools until after a VM has been rebooted, missing a
critical opportunity to configure an instance before or during the
first boot. In order to allow VMWare applications (include vCloud)
to be aware that vm-tools are installed, we need to add the appropriate
metadata to the image during its creation

### DIFF
--- a/modules/KIWIImageFormat.pm
+++ b/modules/KIWIImageFormat.pm
@@ -398,6 +398,11 @@ sub createVMDK {
         $kiwi -> failed ();
         return;
     }
+    my $tools_install_type = 4;
+    my $tools_version = 9344;
+    $this -> __update_vmdk_descriptor(
+        $target, $tools_version, $tools_install_type
+    );
     $kiwi -> done ();
     return $target;
 }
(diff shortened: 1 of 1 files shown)


schaefi:
I did not find a good way to provide the right information for ddb.toolsInstallType and ddb.toolsVersion
Thus these values are currently hard coded

schaefi:
I have tested the result vmdk with qemu and it still boots, however I have not tested this image in vcloud air because I have no access to it

schaefi:
@bear454 if you find time to review this after susecon it would be great, thanks

schaefi:
Thanks for the information, but I think I'll better add a new attribute

bear454:
I disagree: the manual attribute would have to be updated with each update of the open-vm-tools package, and that's a maintenance nightmare considering an environment like OBS.

schaefi:
given you set a version and it's not the same as the installed package provides, how does VMware perform the check whether tools are installed or needs to be updated ?

schaefi:
I really doubt they look into the rpm database on the host, so how does it matter ?

rjschwei:
Robert Schweikert                   MAY THE SOURCE BE WITH YOU
Public Cloud Architect                         LINUX
rjschwei@suse.com 
IRC: robjo




>>> Marcus Schäfer <notifications@github.com> 11/16/15 11:13 AM >>>
given you set a version and it's not the same as the installed package provides, how does VMware perform the check whether tools are installed or needs to be updated ?
Remember that this "check" is not a check for consistency. What "usually" happens is that the tools version is written to the disk when the tools "run" at install time. Thus a mismatch of version number on the disk vs. installed version can, under "normal" circumstances not happen. The issue only exists in "our" model where writing the tools version is performed by a separate process, kiwi, than the tools. Yes, it is possible to write the "wrong" version of the tools to the disk in the "normal" use case as well, but it requires extra effort. Calculating the version based on the description provided by James appear to be a reasonable approach while reducing the risk of mismatch. If we make the user encode the version in the XML file we inevitably will have a problem, especially when building in the build service.

---

bear454:
Looks good to me!

PR #484, opened 2015-09-03
Remove genisoimage and cdrkit-cdrtools-compat dependency (second try)

jotrk:
Based on @schaefi's  last [comment](https://github.com/openSUSE/kiwi/pull/481#issuecomment-134590929) I've updated the pull request with checks to ensure whether `/usr/bin/isoinfo` or `/usr/lib/genisoimage/isoinfo` should be used.

PR #509, opened 2015-10-30
Remove genisoimage and cdrkit-cdrtools-compat dependency (third try)

jotrk:
Follow-up to #484.
The kiwi rpm is built [here](https://build.opensuse.org/package/show/home:jkeil:terminate_wodim/kiwi). The submit request for zisofs-tools, which resolves the Factory issue can be found [here](https://build.opensuse.org/request/show/341666).

### DIFF
--- a/doc/devdoc/devDocs/KIWIIsoLinux.html
+++ b/doc/devdoc/devDocs/KIWIIsoLinux.html
@@ -193,7 +193,7 @@ Called in
 <a name="new"><h2>new</h2></a>
 <p>
      Create a new KIWIIsoLinux object which is used to wrap
-     around the major genisoimage/mkisofs call. This code requires a
+     around the major mkisofs call. This code requires a
      specific source directory structure which is:
 <p>
 Defined on line: 46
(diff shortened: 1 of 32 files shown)


schaefi:
Thanks for not giving up :)

I see the request for zisofs-tools is still in review. Is it ok for you if we wait until this one is checked in ?

Other than that, did you give it a smoke test on the x86 arch with:

```
kiwi --build suse-13.2-JeOS -d /tmp/mytest --type oem
kiwi --build suse-13.1-JeOS -d /tmp/mytest --type oem
kiwi --build suse-leap-42.1-JeOS -d /tmp/mytest --type oem
```

The last two times we tried to get rid of  genisoimage and cdrkit-cdrtools-compat it felt over when I tried to build these templates. The templates are provided by the kiwi-templates package

I haven't tested for rhel6 and rhel7, ideally they should not be affected by the change

Thanks

jotrk:
Aus gegebenem Anlass muss ich dass hier vorzeitig zu machen. Sorry.

schaefi:
all right, waiting for the fourth try :)

jotrk:
Sorry to say, but there will be no fourth try, ping me on IRC or via mail if you want to know more.

PR #510, opened 2015-11-01
[kiwi] add searching in the parent directory of the current path

mmohring:
Search for device drivers in the parent directory of the current path. Kernel modules packages like openSUSE xen-kmp-default put the drivers there. You can with this patch add them with package name="xen-kmp-default" bootinclude="true"/ in the kiwi files drivers section. Currently they are not searched there.


schaefi:
Thanks looks good to me

Issue #475, opened 2015-07-10
A tool is needed to compatibly put isohybrid images on sticks without wasting space


v-fox:
A proper way (the one that allows to use it for advanced Linux booting and file sharing with random legacy systems) to create a boot sticks with 2 partitions, one for an OS and one for redundant space, is:
1) Calculate the amount of the redundant space ({total space} minus {space that's needed for OS}).
2) Allocate that space and create a fat partition, so stupid Windows® systems could see it (because it only looks for the first partition and in its pathetic filesystems). It's not really have to be first, only it's partition number should be and it [can be done manually](http://unix.stackexchange.com/questions/18752/change-the-number-of-the-partition-from-sda1-to-sda2). I think at least one of Linux msdos partitioning tools (cf/f/s/disk, parted, etc.) allows that but it was too long ago since I did such.
3) Allocate that space, create and prepare proper partition for an OS image.
Or, I suppose, one can imitate with Grub or SysLinux what's [IsoStick](https://isostick.com) is doing. But that's overcomplicated, unreliable and has too much limitations (like a size below 4GB).

I noticed that kiwi has something called `livestick` and there is also [these scripts](https://en.opensuse.org/SDB:Live_USB_stick#Using_live-fat-stick_.28Command_line_or_GUI_way.29)  but they are [doing it wrong](https://github.com/cyberorg/live-fat-stick/issues/18).

schaefi:
You are saying livestick did not work for you ? I have tested it today and also made it more robust in terms of device detection (now based on uuid instead of by-id) and it worked good for me

I wrote the following article how to use it:

https://github.com/openSUSE/kiwi/wiki/Create-live-system-from-data-stick

Let me know if that works for you

v-fox:
I don't say "it didn't work", i would say that it's an undocumented thing of unknown purpose that I'm afraid of launching. There is 2 approaches of making a bootable stick that I know: one is described above and chainloading through GRUB or SysLinux. And the latter is more like a software imitation of what [IODD/Zalman SATA boxes](http://www.ixbt.com/storage/esata-p22-zalman.shtml) (for some reason all English mentions of them have disappeared from [Zalman site](http://www.zalman.ru/global/product/product_read.php?id=186) and IODD site is down) are doing. I really would like have an automated way of doing it by the first approach, avoiding chainloading and FAT.

[...]

PR #511, opened 2015-11-02
Working hybrid persistency

v-fox:
These changes supposed to fix critical issues from #475, namely size glitch and loop-file creation loop.
They add new options:
* 'exfat' parameter for 'kiwi_hybridpersistent_filesystem' which requires 'exfat-utils' package and exfat module in initrd (exfat-fuse is not tested)
* 'kiwi_hybridpersistent_filename' for custom loop-file name ('!Live OS's persistent storage.fs' by default)
* 'kiwi_hybridpersistent_filesize' for custom loop-file size which is the half of partition size by default on everything except pure FAT (which is limited to 4G)

Also, for some reason it's not BTRFS which is created in loop-file but the same thing as for RAM-drive. For that reason filesystem for both RAM-drive and loop-file is changes from ext3 to ext4 with journalling disabled. BTRFS was tried but dropped because of kernel panic ("BUG: unable to handle kernel NULL pointer dereference") on kernel 4.2.

A lot of shielding is added into variable assignment for variables that can potentially hold names with spaces and "special character" like the default name for loop-file. It's probably quite an overkill.

Later improvements can be:
* LUKS support
* U/EFi support
* filesystem type selection for RAM and loop-file
* filesystem type selection for persistent storage without Windows® compatibility
* optional compression support for BTRFS used in any manner

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -6473,9 +6473,9 @@ function setupReadWrite {
                     fi
                 fi
             fi
-        # umount the file from read-only mode then finished
-        umount $rwDevice
         fi
+        # umount the file from read-only mode when finished
+        umount $rwDevice
     fi
     return 0
 }
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -23,6 +23,7 @@ export ELOG_FILE=/var/log/boot.kiwi
 export TRANSFER_ERRORS_FILE=/tmp/transfer.errors
 export UFONT=/usr/share/fbiterm/fonts/b16.pcf.gz
 export CONSOLE_FONT=/usr/share/kbd/consolefonts/default8x16.gz
+export HYBRID_PERSISTENT_FILENAME="!Live OS's persistent storage.fs"
 export HYBRID_PERSISTENT_FS=btrfs
 export HYBRID_PERSISTENT_ID=83
 export HYBRID_PERSISTENT_DIR=/read-write
@@ -3925,11 +3926,12 @@ function probeFileSystem {
         squashfs)    FSTYPE=squashfs ;;
         luks)        FSTYPE=luks ;;
         crypto_LUKS) FSTYPE=luks ;;
-        vfat)        FSTYPE=vfat ;;
         clicfs)      FSTYPE=clicfs ;;
         xfs)         FSTYPE=xfs ;;
         udf)         FSTYPE=udf ;;
         zfs_member)  FSTYPE=zfs ;;
+        exfat)       FSTYPE=exfat ;;
+        vfat)        FSTYPE=vfat ;;
         *)
             FSTYPE=unknown
         ;;
@@ -6415,37 +6417,64 @@ function kiwiMount {
 #--------------------------------------
 function setupReadWrite {
     # /.../
-    # check/create read-write filesystem used for
-    # overlay data
+    # check/create read-write ext4 filesystem used for
+    # overlay data in the loop-file on persistency partition
     # ----
     local IFS=$IFS_ORIG
     local rwDir=/read-write
     local rwDevice=`echo $UNIONFS_CONFIG | cut -d , -f 1`
     mkdir -p $rwDir
+    # trying to mount existing file
     if [ $LOCAL_BOOT = "no" ] && [ $systemIntegrity = "clean" ];then
+        # if mount fails, force-check it and try to mount again
         if [ "$RELOAD_IMAGE" = "yes" ] || \
             ! mount -o ro $rwDevice $rwDir &>/dev/null
         then
-            local hybrid_fs=$HYBRID_PERSISTENT_FS
-            if [ ! -z "$kiwi_hybridpersistent_filesystem" ];then
-                hybrid_fs=$kiwi_hybridpersistent_filesystem
+            # store old FSTYPE value
+            if [ ! -z "$FSTYPE" ];then
+                FSTYPE_SAVE=$FSTYPE
             fi
-            Echo "Checking filesystem for RW data on $rwDevice..."
-            checkFilesystem $rwDevice
-
+            # probe and check filesystem
+            probeFileSystem $rwDevice
+            if [ ! "$FSTYPE" = "unknown" ];then
+                Echo "Checking filesystem for RW data on $rwDevice..."
+                # we want to force checking on filesystem without journal
+                fsck -C -y $( [ "$kiwi_hybridpersistent" = "true" ] && echo "-f") $rwDevice
+            fi
+            # restore FSTYPE
+            if [ ! -z "$FSTYPE_SAVE" ];then
+                FSTYPE=$FSTYPE_SAVE
+            fi
+            # if mounting fails anyway, format the file
             if [ "$RELOAD_IMAGE" = "yes" ] || \
                 ! mount -o ro $rwDevice $rwDir &>/dev/null
             then
-                Echo "Creating filesystem for RW data on $rwDevice..."
                 local exception_handling="false"
-                if ! createFilesystem $rwDevice "" "" "hybrid" $exception_handling $hybrid_fs; then
-                    Echo "Failed to create ext3 filesystem"
-                    return 1
+                Echo "Creating filesystem for RW data on $rwDevice..."
+                if [ "$kiwi_hybridpersistent" = "true" ];then
+                    # on flash we want speed & durability over safety
+                    # this is optimized for 512kB erase block size
+                    if ! createFilesystem \
+                        "$rwDevice" "" "" "persistency" "$exception_handling" "ext4" "-F -b 4096 -O ^has_journal -E stride=128,stripe-width=128"
+                    then
+                        Echo "Failed to create hybrid persistency RW filesystem"
+                        return 1
+                    fi
+                else
+                    # this code should not be triggered by anything
+                    # but if it is, let's safely make it work
+                    if ! createFilesystem \
+                        "$rwDevice" "" "" "phantom" "$exception_handling" "ext4"
+                    then
+                        Echo "Failed to create RW filesystem"
+                        return 1
+                    else
+                        Echo "WARNING: RW filesystem successfully created where it shouldn't !"
+                    fi
                 fi
-                checkFilesystem $rwDevice >/dev/null
             fi
-        else
-            umount $rwDevice
+        # umount the file from read-only mode then finished
+        umount $rwDevice
         fi
     fi
     return 0
@@ -6565,7 +6594,9 @@ function mountSystemUnionFS {
         if [ ! "$roDevice" = "nfs" ] && ! setupReadWrite; then
             return 1
         fi
-        if ! kiwiMount "$rwDevice" "$rwDir";then
+       # on flash & RAM we want speed & durability over safety
+        local mount_options="-o defaults,async,relatime,nodiratime,barrier=1"
+        if ! kiwiMount "$rwDevice" "$rwDir" "$mount_options";then
             Echo "Failed to mount read/write filesystem"
             return 1
         fi
@@ -9113,8 +9144,8 @@ function createHybridPersistent {
         local label=$(blkid $partd -s LABEL -o value)
         if [ "$label" = "hybrid" ];then
             Echo "Existing persistent hybrid partition found"
-            if [ $hybrid_fs = "fat" ];then
-                if ! setupHybridCowDevice;then
+            if [ "$hybrid_fs" = "fat" ] || [ "$hybrid_fs" = "exfat" ];then
+                if ! setupHybridCowDevice $partd;then
                     Echo "Failed to setup hybrid cow device"
                     Echo "Persistent writing deactivated"
                     unset kiwi_hybridpersistent
@@ -9151,7 +9182,7 @@ function createHybridPersistent {
         # in bios mode. The partition table created via isohybrid starts at
         # the second partition to allow the creation of the write partition
         # as first partition. Reason for this is to support Windows systems
-        # with a fat partition as write space which has to be the first
+        # with a ex/fat partition as write space which has to be the first
         # partition otherwise Windows can't cope with it.
         #
         # Such a write partition can be created using fdisk, however for
@@ -9161,10 +9192,11 @@ function createHybridPersistent {
         # Therefore we use fdisk for bios firmware images and parted for
         # efi|uefi firmware images.
         #
-        # This also means we don't support fat
+        # This also means we don't support ex/fat
         # based persistent write partitions to be created as first partition
         # on efi|uefi ISO hybrid images
         echo -e "n\np\n$pID\n\n\nw\nq" | fdisk $imageDiskDevice
+        blockdev --rereadpt $device
     else
         createPartitionerInput \
             n p:lxrw $pID . . t $pID $HYBRID_PERSISTENT_ID
@@ -9190,8 +9222,14 @@ function createHybridPersistent {
     #--------------------------------------
     local hybrid_device=$(ddn $device $pID)
     local exception_handling="false"
+    local hybridfs_options=""
+    if [ "$hybrid_fs" = "ext4" ]; then
+        # on flash we want speed & durability over safety
+        # this is optimized for 512kB erase block size
+        hybridfs_options="-b 4096 -O ^has_journal -E stride=128,stripe-width=128"
+    fi
     if ! createFilesystem \
-        $hybrid_device "" "" "hybrid" $exception_handling $hybrid_fs
+        $hybrid_device "" "" "hybrid" $exception_handling $hybrid_fs "$hybridfs_options"
     then
         Echo "Failed to create hybrid persistent filesystem"
         Echo "Persistent writing deactivated"
@@ -9201,12 +9239,12 @@ function createHybridPersistent {
     #======================================
     # export read-write device name
     #--------------------------------------
-    if [ $hybrid_fs = "fat" ];then
-        # The fat filesystem is not really suitable to be used as rootfs
-        # for linux. Therefore we create a btrfs based file which we store
-        # on the fat filesystem and loop setup it. The size of the file
-        # is set to half the size of the fat device
-        if ! setupHybridCowDevice $hybrid_device;then
+    if [ "$hybrid_fs" = "fat" ] || [ "$hybrid_fs" = "exfat" ];then
+        # The ex/fat filesystem is not really suitable to be used as rootfs
+        # for linux. Therefore we create a ext4-based file which we store
+        # on the ex/fat filesystem and loop setup it. The size of the file
+        # is set to half the size of the fat device by default
+        if ! setupHybridCowDevice "$hybrid_device";then
             Echo "Failed to setup hybrid cow device"
             Echo "Persistent writing deactivated"
             unset kiwi_hybridpersistent
@@ -9225,28 +9263,38 @@ function createHybridPersistent {
 #--------------------------------------
 function setupHybridCowDevice {
     local IFS=$IFS_ORIG
-    local hybrid_device=$1
+    local hybrid_device="$@"
+    local hybrid_filename="/cow/${HYBRID_PERSISTENT_FILENAME}"
+    if [ ! -z "$kiwi_hybridpersistent_filename" ];then
+        hybrid_filename="/cow/${kiwi_hybridpersistent_filename}"
+    fi
     mkdir -p /cow
     for i in 1 2 3;do
         mount -L hybrid /cow && break || sleep 2
     done
+    mountpoint -q /cow
     if [ ! $? = 0 ];then
         Echo "Failed to mount hybrid persistent filesystem !"
         return 1
     fi
-    if [ ! -e "/cow/cowfile" ];then
-        local cowsize=$(($(blockdev --getsize64 $hybrid_device) / 2))
-        local exception_handling="false"
-        qemu-img create /cow/cowfile $cowsize
-        if ! createFilesystem \
-            "/cow/cowfile" "" "" "" $exception_handling "btrfs"
-        then
-            Echo "Failed to create hybrid persistent cow filesystem"
-            return 1
+    if [ ! -e "$hybrid_filename" ];then
+        # default filesize is half of partition's capacity
+        local cowsize="$(($(blockdev --getsize64 $hybrid_device) / 2))"
+        # but not for FAT
+        if [ $hybrid_fs = "fat" ];then
+                if [ $cowsize -gt 4294967295 ];then
+                        # filesize is reduced to its limitation
+                        cowsize=4294967295
+                fi
         fi
+        # user can override it on boot
+        if [ ! -z "$kiwi_hybridpersistent_filesize" ];then
+                cowsize=$kiwi_hybridpersistent_filesize
+        fi
+        qemu-img create "$hybrid_filename" $cowsize
     fi
-    export HYBRID_RW=$(loop_setup /cow/cowfile)
-    if [ ! -e "$cowdevice" ];then
+    export HYBRID_RW=$(loop_setup "${hybrid_filename}")
+    if [ ! -e "$HYBRID_RW" ];then
         Echo "Failed to loop setup hybrid cow file !"
         return 1
     fi
@@ -9898,13 +9946,17 @@ function createFilesystem {
             local bytes=$((blocks * 4096))
             opts="$opts -b $bytes"
         fi
+    elif [ "$filesystem" = "xfs" ];then
+        if [ ! -z "$label" ];then
+            opts="$opts -L $label"
+        fi
     elif [ "$filesystem" = "fat" ];then
         if [ ! -z "$label" ];then
             opts="$opts -n $label"
         fi
-    elif [ "$filesystem" = "xfs" ];then
+    elif [ "$filesystem" = "exfat" ];then
         if [ ! -z "$label" ];then
-            opts="$opts -L $label"
+            opts="$opts -n $label"
         fi
     elif [ "$filesystem" = "ntfs" ];then
         if [ ! -z "$label" ];then
@@ -9932,6 +9984,8 @@ function createFilesystem {
         fi
     elif [ "$filesystem" = "fat" ];then
         mkfs.fat $opts $deviceCreate $blocks 1>&2
+    elif [ "$filesystem" = "exfat" ];then
+        mkfs.exfat $opts $deviceCreate 1>&2
     elif [ "$filesystem" = "ntfs" ];then
         mkfs.ntfs $opts $deviceCreate $blocks 1>&2
     else
@@ -11384,12 +11438,12 @@ function backupGPT {
 #--------------------------------------
 function loop_setup {
     local IFS=$IFS_ORIG
-    local target=$1
+    local target="$@"
     local logical_sector_size
     if [ ! -z "$kiwi_target_blocksize" ];then
         logical_sector_size="-L $kiwi_target_blocksize"
     fi
-    local loop=$(losetup $logical_sector_size -f --show $target)
+    local loop=$(losetup $logical_sector_size -f --show "$target")
     if [ ! -e "$loop" ];then
         return 1
     fi
@@ -11400,8 +11454,8 @@ function loop_setup {
 #--------------------------------------
 function loop_delete {
     local IFS=$IFS_ORIG
-    local target=$1
-    losetup -d $target
+    local target="$@"
+    losetup -d "$target"
 }
 #======================================
 # startMultipathd
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -6548,6 +6548,7 @@ function mountSystemUnionFS {
     local roDevice=`echo $UNIONFS_CONFIG | cut -d , -f 2`
     local unionFST=`echo $UNIONFS_CONFIG | cut -d , -f 3`
     local prefix=/mnt
+    local rwMountOptions=""
     #======================================
     # load fuse module
     #--------------------------------------
@@ -6594,9 +6595,13 @@ function mountSystemUnionFS {
         if [ ! "$roDevice" = "nfs" ] && ! setupReadWrite; then
             return 1
         fi
-       # on flash & RAM we want speed & durability over safety
-        local mount_options="-o defaults,async,relatime,nodiratime,barrier=1"
-        if ! kiwiMount "$rwDevice" "$rwDir" "$mount_options";then
+        if [ "$kiwi_hybridpersistent" = "true" ];then
+            # on flash we want speed & durability over safety.
+            # these options may be usefull for everyone but
+            # we'll abstain from forcing them elsewhere
+            rwMountOptions="-o defaults,async,relatime,nodiratime"
+        fi
+        if ! kiwiMount "$rwDevice" "$rwDir" "$rwMountOptions";then
             Echo "Failed to mount read/write filesystem"
             return 1
         fi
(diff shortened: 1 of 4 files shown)


schaefi:
Thanks much for your effort. I'll look into the code now. On a first quick check it looks good to me. Would you mind to squash your commits into the following categories:

* bash style fixes, quotation
* exfat support, replacing my simple fat code
* changes to the schema and corresponding code

Thanks, now looking into the code

modules/KIWILinuxRC.sh
```diff
@@ -23,6 +23,7 @@ export ELOG_FILE=/var/log/boot.kiwi
 export TRANSFER_ERRORS_FILE=/tmp/transfer.errors
 export UFONT=/usr/share/fbiterm/fonts/b16.pcf.gz
 export CONSOLE_FONT=/usr/share/kbd/consolefonts/default8x16.gz
+export HYBRID_PERSISTENT_FILENAME="!Live OS's persistent storage.fs"
```
schaefi: That's a strange name for linux users :-) I guess you use that to make it more clear on the Windows side when Windows provides a view on the iso ? if so a comment would be nice.

modules/KIWILinuxRC.sh
```diff
@@ -6420,17 +6422,26 @@ function setupReadWrite {
...
             if [ "$RELOAD_IMAGE" = "yes" ] || \
-                ! mount -o ro $rwDevice $rwDir &>/dev/null
+                ! mount -o ro "$rwDevice" $rwDir &>/dev/null
             then
-                Echo "Creating filesystem for RW data on $rwDevice..."
-                if ! mkfs.ext3 -F $rwDevice >/dev/null;then
-                    Echo "Failed to create ext3 filesystem"
-                    return 1
+		Echo "Creating filesystem for RW data on $rwDevice..."
+                if [ "$kiwi_hybridpersistent" = "true" ];then
```
schaefi: The indentation is broken here and in other places, could you fix that (4spaces) thanks

modules/KIWILinuxRC.sh
```diff
@@ -9219,19 +9236,24 @@ function setupHybridCowDevice {
...
+        local cowsize="$(($(blockdev --getsize64 $hybrid_device) / 2))"
+        # but not for FAT
+        if [ $hybrid_fs = "fat" ];then
+	        if [ $cowsize -gt "4294967295" ];then
+		        # filesize is reduced to its limitation
+		        cowsize=4294967295
+	        fi
+        fi
+        # user can override it on boot
+	if [ ! -z "$kiwi_hybridpersistent_filesize" ];then
```
schaefi: maybe I overlooked it, but I see you have changed the schema to provide an attribute for this but I don't see the code evaluating the XML and adding kiwi_hybridpersistent_filesize as part of the profile environment so that it can be actually read

modules/KIWILinuxRC.sh
```diff
@@ -9888,6 +9914,8 @@ function createFilesystem {
         xfs_admin -U $uuid $deviceCreate
     elif [ "$filesystem" = "fat" ];then
         mkfs.fat $opts $deviceCreate $blocks 1>&2
+    elif [ "$filesystem" = "exfat" ];then
+        mkfs.exfat $opts $deviceCreate 1>&2
```
schaefi: mkfs.exfat is required to be part of the initrd, but it's not listed to stay there. See metadata/KIWIConfig.xml
imho we need an entry:

<file name="mkfs.exfat"/>

in the tools section

modules/KIWISchema.rnc
```diff
@@ -1783,8 +1783,20 @@ div {
...
         attribute hybridpersistent_filesystem {
-            "btrfs" | "fat"
+            "btrfs" | "fat" | "exfat"
         }
+    k.type.hybridpersistent_filename.attribute =
+        ## for the iso type only:
+        ## Set the filename to use for persistency file if ex/fat is used for
+        ## hybrid ISO write support. Default is "!Live OS's persistent storage.fs"
+        attribute hybridpersistent_filename { text }
+    k.type.hybridpersistent_filesize.attribute =
```
schaefi: both attributes will not find their way into the initrd profile environment, maybe this was just forgotten ?

modules/KIWILinuxRC.sh
v-fox: > That's a strange name for linux users :-) I guess you use that to make it more clear on the Windows side when Windows provides a view on the iso ? if so a comment would be nice.

Yeah, it so someone (who is most likely a clueless Windows user) looking at that partition wouldn't get an idea of deleting this file. And to test compatibility with "weird" names.

modules/KIWILinuxRC.sh
v-fox: > The indentation is broken here and in other places, could you fix that (4spaces) thanks

Yeah, I usually use 8 tabs instead of 4 spaces and when I figured that it must be breaking your style it was too late.

modules/KIWILinuxRC.sh
v-fox: > maybe I overlooked it, but I see you have changed the schema to provide an attribute for this but I don't see the code evaluating the XML and adding kiwi_hybridpersistent_filesize as part of the profile environment so that it can be actually read

I've edited KIWISchema.rnc to add those and it seem to work (I've tested custom names and sizes on isolinux's boot prompt) but I don't know what else should be changed and why.

modules/KIWILinuxRC.sh
v-fox: > mkfs.exfat is required to be part of the initrd, but it's not listed to stay there. See metadata/KIWIConfig.xml
> imho we need an entry:

Indeed, I wasn't even aware of that file and decided not to cluelessly try to add a dependency somewhere.

modules/KIWISchema.rnc
v-fox: > both attributes will not find their way into the initrd profile environment, maybe this was just forgotten ?

I have no idea what that means :] But those parameters do work.

v-fox:
> Would you mind to squash your commits into the following categories:
> 
>     bash style fixes, quotation
>     exfat support, replacing my simple fat code
>     changes to the schema and corresponding code
> 
> Thanks, now looking into the code

Well, I'm not sure on how to do it right. This is literally my first pull request and I'm glad that I've managed to do it not in one big glob of data at all.
I wouldn't call anything there "a replacement of simple fat code", it's more of an addition. It's one particular commit (5a18da55fc9a30259091d00e2f8890c921c15237) which is full of shielding changes and the biggest one. Most of them are probably unnecessary and the only real fix it provides is returning "/cow" directory back into place where it should be. I don't quite remember, these changes have accumulated in about a month or so. I didn't plan so many of them at the start.

Check [this patched kiwi build on OBS](https://build.opensuse.org/package/show/home:X0F:HSF/kiwi). And soon I'll make a public build with these patches.

I may try to fix my mishaps latter but I'm afraid to break something while doing so.

schaefi:
you can call

```
git rebase -i
```

on your branch. In the following editor view you will see your 12 commits, you can now squash, reword them. There is a comment written in the editor view how to do it. Once done you can call

```
git push --force
```

That automatically changes the view for me here and if we merge it, it will have a better structure of what has changed.

v-fox:
* updated
* fixed the indentations
* removed all unnecessary shielding ("bash style fixes, quotation") that was not pertaining to persistency feature  
(I think, needs testing)
* added <file name="mkfs.exfat"/> dependency
(but for feature to work, exfat kernel module or, possibly, exfat-fuse should be in initrd)

— I still don't think that I actually "replaced" your FAT code, FAT should work, just with 4GB limit
— I still don't understand what "changes to the schema and corresponding code" and "code evaluating the XML and adding <X> as part of the profile environment" are exactly. I've tested changing those variables on bootloader's command line but not in kiwi's xml config. So, maybe something else should be tweaked somewhere for them to work there.

schaefi:
If you squash your commits the result should be the three commits I suggested

* bash style fixes, quotation
* exfat support, replacing my simple fat code
* changes to the schema and corresponding code

These were the three main areas touched and which could be handled and merged conflict free
I can jump in and fix the missing parts of the schema changes you have added but I can only do that once the pull request is in a shape to allow an intermediate change or commit based review

actually the change set is imho too big already and it's generally not a good idea to mix many changes into one big blob. However I can understand that this happens if people just want to get something done, but you are touching boot code and that's a critical part where we need to be really careful

Hope this makes sense to you too.

Alternatively you can close this one and create new pull requests, starting with e.g only the bash style fixes. We would be able to merge that in quickly. As result the change set will become smaller.
Next you can come up with your exfat additions in a way that it is not directly connected to an additional schema change. And last but not least we could add the schema change to make it configurable together or you can let me do it if preferred

I'm afraid I can't merge in 17 commits of basically random changes, if I encounter that something is broken I would have a very hard time to find out which one of these might have caused it.

I'm sorry

v-fox:
> If you squash your commits the result should be the three commits I suggested
> 
>     bash style fixes, quotation
>     exfat support, replacing my simple fat code
>     changes to the schema and corresponding code

I noticed that and answered already:
* I [removed all "bash style fixes, quotation"](https://github.com/v-fox/kiwi/commit/593a584560ac1d6b8214b2ba880fc3e666a485ac) that are not directly pertaining to the feature. Now it's about twice as small. [See](https://github.com/openSUSE/kiwi/pull/511/files), that shouldn't be excessive.
But I haven't tested it on real hardware yet. I will try in the next 24 hours.
* That's the meat of it.
* But I have no idea what "schema and corresponding code" exactly is. If its changes to modules/KIWISchema.rnc, then it doesn't make much sense to divide them since options are introduced simultaneously to boot code and there. It can be easily done and default values would be used in boot-code. But those are pretty minor.
BUT I don't know if something else should be changed, so they would be compliant with your rules. So, what exactly is missing ?

> actually the change set is imho too big already and it's generally not a good idea to mix many changes into one big blob. However I can understand that this happens if people just want to get something done, but you are touching boot code and that's a critical part where we need to be really careful

Yeah, I myself is wary of that. But boot code changes is what it's all about.

> Alternatively you can close this one and create new pull requests, starting with e.g only the bash style fixes. We would be able to merge that in quickly. As result the change set will become smaller.

I started those "bash style fixes, quotations" to make sure that filenames with spaces are not breaking in functions. But at the end I've noticed that you've used IFS redefinition in places where it might happen. So I removed it now. It should be safe but you should test it on your end too before accepting, to be sure.

> Next you can come up with your exfat additions in a way that it is not directly connected to an additional schema change. And last but not least we could add the schema change to make it configurable together or you can let me do it if preferred

Those changes to modules/KIWISchema.rnc should be safely removable and default values should be used. I just don't understand what's wrong about it. What exactly is it missing ?

> I'm afraid I can't merge in 17 commits of basically random changes, if I encounter that something is broken I would have a very hard time to find out which one of these might have caused it.
> 
> I'm sorry

I can squash it. I did it to make [one patch for testing here](https://build.opensuse.org/package/view_file/home:X0F:HSF/kiwi/kiwi-hybrid_persistency.patch?expand=1). But before I'll do it, I would like to get your approval of those changes, as it would be effectively a finalization. It should be already viewable as one complete change in here.

v-fox:
Goddamit ! If you are not going to merge it, could you at least not to break it more ? It's called "Working hybrid persistency" for a reason, because it fixes errors in both logic and implementation in your unfinished code and adds missing basic functions.

Then I finish testing it, **actually testing it... for working as intended**, I will squash it in one single commit.

schaefi:
Well blaming on me doesn't help. I did not break anything related to this pull request and I'm not going to stop development and other people's contributions just because there is a pending pull request which doesn't care for the review suggestions I have made.

I'm sorry if you squash 23 commits with different scopes in different areas of the code into one I'm not going to merge it. You might also consider that you are not the only one who has expectations to the functionality and I'm pretty sure you have not tested anything else outside of your use case.

I run an integration test system which tests many different use cases and if only one is broken by the stuff added here I currently won't have a good chance to find the reason why in a fast way.

I already told you that your code changes covers three basic areas:

* bash style fixes, quotation
* exfat support, replacing my simple fat code
* changes to the schema and corresponding code

If you can't structure the changes in a way that we can maintain it I'm sorry. If you would have considered to just close this request and create three new ones we would have already achieved something

> It's called "Working hybrid persistency" for a reason, because it fixes errors in both logic and
> Implementation in your unfinished code and adds missing basic functions.

The current implementation works, you are trying a different approach with exfat which might make it better. Something like this is called refactoring and not adding missing functionality. However the code I have reviewed here is definitly incomplete concerning the schema additions, it's missing the data handling methods. yet another reason to break it up into several pull requests

Last but not least if I'm allowed to give an advise; It's always better to prevent personal attacks when talking about technology. No idea what or where you work but living an offending style always causes trouble and does not solve anything

v-fox:
> Well blaming on me doesn't help. I did not break anything related to this pull request and I'm not going to stop development and other people's contributions just because there is a pending pull request which doesn't care for the review suggestions I have made.

You did not break it in your code because it doesn't work without my changes in the first place.
What you have to back up your allegation about me "not caring" ? I answered to you that you didn't not understand how this code works... even thought you wrote half of it. I explained every change and asked you questions about your vague demands which you ignored. Am I really the one not caring ?

setupReadWrite function doesn't even format persistency partition, **you've yourself put that into createHybridPersistent** ! So why now you've merged code that tries to use persistency variable in the wrong function but ignored my fixes that use it correctly ?

> I'm sorry if you squash 23 commits with different scopes in different areas of the code into one I'm not going to merge it.

Well, fixing your incomplete implementation requires changes in all of its places. Why do you think it's OK for you to drop broken code but it's bad for me to fix it just because it's more than a few lines ? First you've said that "I missed something" then it's too many "different areas", the opposite.

> You might also consider that you are not the only one who has expectations to the functionality and I'm pretty sure you have not tested anything else outside of your use case.

My case is pretty encompassing but my changes should not affect much outside of their purpose. You for some reason are making changes to this code that does not fix critical issues in this feature and doesn't change anything for anything else. You're "fixing" parts of the code that work without touching the broken parts.

> I run an integration test system which tests many different use cases and if only one is broken by the stuff added here I currently won't have a good chance to find the reason why in a fast way.

Did you even actually booted a system with a RW overlay loop-file on a FAT partition ?

> I already told you that your code changes covers three basic areas:
> 
>     bash style fixes, quotation
>     exfat support, replacing my simple fat code
>     changes to the schema and corresponding code

And I told you that **it's not**.
1) I removed all those "bash style fixes, quotation". EVERY. SINGLE. ONE.
2) It doesn't replace all of you code. It replaces broken parts and expanding the working ones. Slightly. It doesn't "replace fat code for exfat", it removes FAT hardcoding to make it filesystem-agnostic and allows formatting and mounting in exfat.
What it really replaces is your ext3 code from setupReadWrite to ext4 because that function formats loop-file on persistency partition if it itself formatted with Windows filesystem.
It should be expanded in the future to also be filesystem-agnostic and user option should be provided. BUT it used ONLY in the loop-file... it also was used by ramdrive until [you removed it](). It maybe used somewhere else but now that use case also is broken because the function is broken. By you. Like [you broke third-party modules in initrd](https://github.com/openSUSE/kiwi/commit/ea315c4a2e00aec10ad1644885ad1e2197794325) recently and then [silently but incompletely](https://github.com/openSUSE/kiwi/commit/09c410bccb567a06fa975844124e61b011086173) fixed them ("extra" directory is still removed and kiwi refuses to not strip, ignoring user choice).
3) And how am I supposed to add new variables for name and size of the file without allowing them to be used by kiwi ? You yourself previously wrote that "both attributes will not find their way into the initrd profile environment, maybe this was just forgotten ?". Well... should I make them work as intended or should I leave them for no reason ?
I can remove them... but what's the point ? What's your problem with them ?

> If you can't structure the changes in a way that we can maintain it I'm sorry. If you would have considered to just close this request and create three new ones we would have already achieved something

And how "the new" request be any different ? This is how patches are, I may change them, add or remove them or merge them. What the hell do you think I should do ? Merge 23 into 3 but not 1 because you like number 3 more than 1 ?
Changes from your 1st demand don't even exist and 3rd works in conjunction with 2. So, I should break the function just for the sake of breaking it and then fixing it separately ?

>     It's called "Working hybrid persistency" for a reason, because it fixes errors in both logic and
>     Implementation in your unfinished code and adds missing basic functions.
> 
> The current implementation works, you are trying a different approach with exfat which might make it better. Something like this is called refactoring and not adding missing functionality. However the code I have reviewed here is definitly incomplete concerning the schema additions, it's missing the data handling methods. yet another reason to break it up into several pull requests

It is refactoring. But again, did you even actually booted a system with a RW overlay loop-file on a FAT partition ? Because then I used your code as it is, [it made broken loop-file](https://github.com/openSUSE/kiwi/issues/475#issuecomment-126405087) (even worse than not to make one at all) and system failed to properly boot. And you ignored all my complains and reports. Which is why I had no choice but to fix it myself.

And I spend months on your code ! And no one compensated my time. It's fine since I'm the one who requested that function. But if you think that it lacks something which I have no idea about, maybe you better off adding it as you think it's should be or redoing everything yourself. And properly this time.

> 
> Last but not least if I'm allowed to give an advise; It's always better to prevent personal attacks when talking about technology. No idea what or where you work but living an offending style always causes trouble and does not solve anything

It also better to not ignore what is said to you and think about what it is instead of being fanatically stuck with some preconception. In fact, it's even more offensive and disrespectful than any kind of most vicious direct attack.
Don't think that you're a nice guy only because you use nice word. It's not the words that matter. That goes for whole current SUSE team.

v-fox:
Here I remade all those commits into 2, script changes and "metadata" changes, by squashing (should have been done anyway because of too many back-and-forth changes in the original) and dividing them manually.

These last changes I didn't test thoroughly, only on their main purpose: exfat loop-file on flash drive. So testing with native Linux filesystems and with persistency off would be useful to be safe. But I do not expect any issues.
Also, it would be advisable to check if anything anywhere besides this loop-file uses setupReadWrite function. It was hardcoded for ext3 until recently which couldn't be preferred by anyone.

PS: And you still didn't answer my same questions which are not rhetorical \>_\<

schaefi:
sorry took a while to come back to this one

> It also better to not ignore what is said to you and think about what
> it is instead of being fanatically stuck with some preconception. In fact,
> it's even more offensive and disrespectful than any kind of most vicio
> us direct attack.

wow you are saying I'm ignoring you, and what I suggested is meant
rhetorical and I'm fanatically stuck with some preconception.
That seems to be a bunch of personal prejudices I did not send
to you, did I ?

Everything I said was meant honest and not rhetorical. I intentionally
choose a respectful conversation in order to prevent people from
feeling attacked even if they attack me. I really wonder why this
conversation went into that direction.

This is all about technology and no matter what vulgarness people
throw on me I won't see it any differently :) I can prove that I do not
disrespect your work, just the contrary because I took your code
and put it into a shape so that I can test and merge it.

This was actually the part I have asked you to do in one of my first
messages after the initial review in this thread. Obviously I was not
able to make my point clear. Therefore I now try to come up with a
pull request based on your work to see if that make sense

In any case what you did would still be a good improvement, here ist
it in a little more structured form and if you are still willing to
review it I would be happy. Nothing I have said is meant rhetorical,
just recommendations. I can't do more than saying it again.

    https://github.com/openSUSE/kiwi/pull/537

I have tested this code with an example based on 13.2. exfat support
is implemented there using fuse, thus code to mount it fuse based
was required too, at least for me.

The two XML options:

* kiwi_hybridpersistent_cow_filename
* kiwi_hybridpersistent_filesize

are not handled so far. We can add that if you are still interested

> did you even actually booted a system with a RW overlay loop-file on 
> a FAT partition ?

yes I have tested that several times in a virtual environment though.
Here are my type definitions:

```
<type image="iso" boot="isoboot/suse-13.2" flags="overlay" hybrid="true" hybridpersistent_filesystem="fat" hybridpersistent="true"/>
```

and with exfat

```
<type image="iso" boot="isoboot/suse-13.2" flags="overlay" hybrid="true" hybridpersistent_filesystem="exfat" hybridpersistent="true"/>

also needed the following packages:

<package name="exfat-utils" bootinclude="true"/>
<package name="fuse-exfat" bootinclude="true"/>
```

And here is the process of testing:

```
qemu-img create mydisk 5g
dd if=LimeJeOS-openSUSE-13.2.x86_64-1.13.2.iso of=mydisk conv=notrunc
qemu-kvm mydisk
```

schaefi:
Also tested with flags="clic"

v-fox:
> sorry took a while to come back to this one

Well, it is a holiday season, the time when it should take a while ;]

> wow you are saying I'm ignoring you, and what I suggested is meant rhetorical and I'm fanatically stuck with some preconception. That seems to be a bunch of personal prejudices I did not send to you, did I ?

Can you honestly say that you actually answered all of my questions ?
You don't, at least, because, as I stated just beforehand, I started this whole "refactoring" after you ignored my questions about current kiwi code being completely unusable even with FAT because it creates 1MB loop-file for some unknown reason on my system. You've just repeated the same demands without answering anything again and again.

And for something to be "prejudice" it should be based on unrelated facts, like you being an arab lesbian ex-Microsoft worker or something. I guess "a SUSE team member" may be a basis of prejudice but only if that status is used for criticism and not actual deeds and words or lack there of.

> Everything I said was meant honest and not rhetorical. I intentionally choose a respectful conversation in order to prevent people from feeling attacked even if they attack me. I really wonder why this conversation went into that direction.

Because my questions are ignored for months. As my attempts to fix stuff on my own.

> This is all about technology and no matter what vulgarness people throw on me I won't see it any differently :) I can prove that I do not disrespect your work, just the contrary because I took your code and put it into a shape so that I can test and merge it.

Since my code was putting-your-code-into-a-shape-in-which-it-works-as-expected, I'm not sure of you redoing it instead of proposing changes to that request in clear and factual manner is respectful or not.

> This was actually the part I have asked you to do in one of my first messages after the initial review in this thread.

And what is "this" exactly ?
Don't forget that then I started to write my changes I didn't know what kind of issues I may encounter, where and why, what functions are called and in which order. And I did only because you didn't answer where I should begin to search for cause of my initial issues.

> I have tested this code with an example based on 13.2. exfat support is implemented there using fuse, thus code to mount it fuse based was required too, at least for me.

Make sure that fuse-exfat is not a hard dependency because it overrides kernel module and slows down speed to a crawl on already slow medium, especially with 2 layers of filesystems.

* You seem to explicitly use `mount.exfat`. **You shouldn't do that**.
`mount` automatically picks it up if fuse-exfat is installed even if kernel module is present, without fuse-exfat `mount` also automatically uses kernel module. Ideally, something should be done for kernel module to take precedence. But your version completely ignores it which is the worst.
* It also seems that you continue to treat setupReadWrite as if it formats RW partition which may result in 2 partition having "hybrid" label, one containing loop-file and loop-file itself. But that probably would only happen in case if loop-file already exists but is not mountable for some reason. Still I'm not sure why duplicate hybrid_fs formatting in both setupReadWrite and createHybridPersistent.
Adding something like this may be warranted:
```
local hybrid_label=hybrid
if [ "$hybrid_fs" = "fat" ] || [ "$hybrid_fs" = "exfat" ];then
    hybrid_label=persistency
fi
if ! createFilesystem \
            "$rwDevice" "" "" "$hybrid_label" "false" "$hybrid_fs" "$fs_opts"
```
* Most likely loop-file with ext4 without journal should be checked with `-f` option because it's "dirty" state may be missed in case of corruption by sudden power loss or unplugging.
You've added the `-f` option but the problem is that it may be mountable while being corrupted therefore you may need to force the check on every boot and before the mount to be sure.
* However, it may leave the user with a black screen because of some kind of race condition with KMS module initialization and kiwi's init that I've encountered a few times. So, if you're pruning "extraneous" modules from the initrd, you may want to remove KMS modules. It also will allow addition of proprietary NVIDIA drivers. But maybe it's already possible, I'm not sure.

That also brings us to you not commenting again about kiwi removing `extra` kernel module folder, `updates` kernel module folder removal mishap that leaved me unable to make working builds with exfat at all for a while and kiwi ignoring changes to `images.sh`.

Your version seems to use ext4 parameters for both loop-file and partition and that is good.
* BUT the fact that setupHybridCowDevice function formats the loop-file without accounting for HYBRID_EXT4_OPTS and without the label (again, anything but "hybrid", like "persistency") isn't. That way, options will never be used on the loop file and it will degrade flash with excessive writes.

You also seem to have adopted mount options of `-o defaults,async,relatime,nodiratime` which is good.
* BUT adding `barrier=1` was a mistake because it's ext4-specific and may not work with xfs or btrfs. It's seems to be a leftover from when I thought that `barrier=0` may be useful for speed but then decided that it's too dangerous (again, in case of power loss or sudden unplugging) while not influencing the amount of writes at all.

> The two XML options:
> 
>     kiwi_hybridpersistent_cow_filename
>     kiwi_hybridpersistent_filesize
> 
> are not handled so far. We can add that if you are still interested

I'm again not sure what you mean by "not handled" because you once again don't explain. But filesize option should be present because it may be very desirable on large-capacity flash drives where big-sized loop-file might be an overkill.

While you at it you may explore the possibility of creating LUKS-encrypted RW partition/loop-file. I think kiwi's init script has functions for opening existing LUKS-partitions but not creating them. openSUSE-based encrypted flash-drives may be more robust and useful than anything that is out there now.

> yes I have tested that several times in a virtual environment though.

But did you test your original code (because of which I even started this all) on real flash drives ?

> And here is the process of testing:
> 
> qemu-img create mydisk 5g
> dd if=LimeJeOS-openSUSE-13.2.x86_64-1.13.2.iso of=mydisk conv=notrunc
> qemu-kvm mydisk

Does this mean that kiwi does not check if it's working with removable media (like `imagewriter` does) ? Isn't it dangerous ?

schaefi:
>    Can you honestly say that you actually answered all of my questions ?
>    You don't, at least, because, as I stated just beforehand, I started
>    this whole "refactoring" after you ignored my questions about current
>    kiwi code being completely unusable even with FAT because it creates
>    1MB loop-file for some unknown reason on my system. You've just
>    repeated the same demands without answering anything again and again.

As far as I know I repeated with, that I can't reproduce this and I still
can't reproduce this behavior. It is correct that it happens that I don't
find the time to answer all questions people sent to me. This is either
because I have other tasks or because I can't say anything useful other
than that I can't be really helpful. The consequence is that people find
alternative solutions or at best start hacking. You started hacking
because you felt ignored ? I'm sorry I don't understand that. I would
feel ignored when I start hacking and nobody would respond to my
contributions. Maybe all this is just a great misunderstanding

In any case I think all changes should be done in small steps and not
everything together. If there are still questions open they will be
answered and addressed in the process. It is not meant as a personal
attack when a question stays open

>    Since my code was
>    putting-your-code-into-a-shape-in-which-it-works-as-expected, I'm not
>    sure of you redoing it instead of proposing changes to that request in
>    clear and factual manner is respectful or not.
> 
>      This was actually the part I have asked you to do in one of my first
>      messages after the initial review in this thread.
> 
>    And what is "this" exactly ?

your code is good, it was just not in the form that it was possible to
commit small junks of changes which do not interfere with each other.
That's what my pull request does and should explain the "this".
Feel free to take over my pull request, you will see that it is your
code just in a bit different form. I know some software developers feels
sore when touching their code but really this is not meant as a personal
affront, it makes things better.

>    Make sure that fuse-exfat is not a hard dependency because it overrides
>    kernel module and slows down speed to a crawl on already slow medium,
>    especially with 2 layers of filesystems.
>      * You seem to explicitly use mount.exfat. You shouldn't do that.

good point

>      * It also seems that you continue to treat setupReadWrite as if it
>        formats RW partition which may result in 2 partition having
>        "hybrid" label, one containing loop-file and loop-file itself.

Hmm, you see this in the clicfs case, right ? I will check that

>      * Most likely loop-file with ext4 without journal should be checked
>        with -f option because it's "dirty" state may be missed in case of
>        corruption by sudden power loss or unplugging. You've added the -f
>        option but the problem is that it may be mountable while being
>        corrupted therefore you may need to force the check on every boot
>        and before the mount to be sure.

ok, that can be added too. However I really hope a "normal" reboot doesn't
leave the write partition in a dirty state.

>      * However, it may leave the user with a black screen because of some
>        kind of race condition with KMS module initialization and kiwi's
>        init that I've encountered a few times. So, if you're pruning
>        "extraneous" modules from the initrd, you may want to remove KMS
>        modules. It also will allow addition of proprietary NVIDIA drivers.
>        But maybe it's already possible, I'm not sure.

ok, can we handle the KMS stuff in another thread once this is done ?

>    That also brings us to you not commenting again about kiwi removing
>    extra kernel module folder, updates kernel module folder removal mishap
>    that leaved me unable to make working builds with exfat at all for a
>    while and kiwi ignoring changes to images.sh.

It might be because this bug was also reported in our bugzilla system
and fixed in commit 09c410bccb.

>      * BUT the fact that setupHybridCowDevice function formats the
>        loop-file without accounting for HYBRID_EXT4_OPTS

yes good point, needs a fix

>    You also seem to have adopted mount options of -o
>    defaults,async,relatime,nodiratime which is good.
>      * BUT adding barrier=1 was a mistake because it's ext4-specific and
>        may not work with xfs or btrfs.

ok let's remove it. By the way originally I had planned to use btrfs
but it does not work the way we use it here. There is a bug against btrfs
on the kernel mailing list. Thus this defect needs a fix in the filesystem
itself. xfs I haven't tested. Problem with inode based filesystems is
that its initial creation might take quite some time

>      The two XML options:
> kiwi_hybridpersistent_cow_filename
> kiwi_hybridpersistent_filesize
> 
>      are not handled so far. We can add that if you are still interested
> 
>    I'm again not sure what you mean by "not handled" because you once
>    again don't explain.

ok, the option in the XML is transfered into a shell profile environment
file called '.profile'. Handling means that the XML option needs to be
written into this file, which is done by KIWIProfileFile.pm. I think this
can be done like KMS in an extra thread once we merged the first round


> But filesize option should be present because it
>    may be very desirable on large-capacity flash drives where big-sized
>    loop-file might be an overkill.

yes agreed, we will do it but let's first fix on the basics

>    While you at it you may explore the possibility of creating
>    LUKS-encrypted RW partition/loop-file. I think kiwi's init script has
>    functions for opening existing LUKS-partitions but not creating them.
>    openSUSE-based encrypted flash-drives may be more robust and useful
>    than anything that is out there now.

kiwi supports the luks encryption at image build time. However with the
persistent writing we moved the creation of the filesystem into the boot
code and there you are right only opening an existing luks is implemented.
This would be yet another new story. It would be a valuable feature yes

>      yes I have tested that several times in a virtual environment
>      though.
> 
>    But did you test your original code (because of which I even started
>    this all) on real flash drives ?

Before you jumped in I used a fat based USB stick with the "fat" filesystem
and overlayfs. I did not use clicfs. So for me things worked

>    Does this mean that kiwi does not check if it's working with removable
>    media (like imagewriter does) ? Isn't it dangerous ?

This test in a virtual environment can be automated and is the first set
of tests. If it works there manual integration tests on USB sticks and SD
cards follows. However the latter is done by community members and they
normally report back when something breaks

Adding your improvement suggestions now

Thanks

schaefi:
> It also seems that you continue to treat setupReadWrite as if it
> formats RW partition which may result in 2 partition having
>  "hybrid" label, one containing loop-file and loop-file itself.

Did some test builds and could not reproduce this problem. Here the result for the following build
type

```
<type image="iso" primary="true" boot="isoboot/suse-13.2" flags="clic" hybrid="true" firmware="bios" kernelcmdline="splash" hybridpersistent_filesystem="exfat" hybridpersistent="true"/>
```

```
blkid /dev/mapper/loop0p1 

/dev/mapper/loop0p1: LABEL="hybrid" UUID="6A10-B930" TYPE="exfat" PTTYPE="dos" PARTUUID="eb3fc47f-01"
```

```
mount /dev/mapper/loop0p1 /mnt
blkid /mnt/Live\ OS\'s\ persistent\ storage.fs 
/mnt/Live OS's persistent storage.fs: UUID="fd8a5a09-62f3-45b1-b6e8-87f739a11702" TYPE="ext4" 
```

Looks good to me

schaefi:
I'm going to merge the derived pull request and will add your name as the originator to the merge commit. I hope this is ok for you. Further patches can be added in additional pull requests. I'm going to hand over the package to the integration testing system

Thanks

PR #515, opened 2015-11-10
Bundle all built OEM images

seife:
it is valid to build multiple OEM targets, e.g. with
&lt;type ... installiso='true' installpxe='true' image='oem'&gt; and
KIWI correctly builds an install.iso and an install.tar.xz in
this case. However, when using OBS only the install.iso is
exported. This is, because "kiwi --bundle-build" will only export
one image and return.
Change this to export all: .iso, .raw and .tar.xz


schaefi:
Thanks, good catch

PR #520, opened 2015-11-20
Add warning to extractkernel  for aarch64/armv7/s390x

dirkmueller:
For those architectures, the kernel links are named differently
than on x86/ppc.

### DIFF
--- a/modules/KIWIImage.pm
+++ b/modules/KIWIImage.pm
@@ -4192,6 +4192,11 @@ sub extractLinux {
         }
         KIWIQX::qxx ("rm -rf $imageTree/boot/*");
     }
+    else {
+        $kiwi -> warning  ("--> Can't find kernel for extraction: " .
+                           "did you call suseStripKernel?");
+        $kiwi -> skipped ();
+    }
     return $name;
 }
 


modules/KIWIImage.pm
```diff
@@ -4125,9 +4125,17 @@ sub extractLinux {
...
-        $kernname = "vmlinux";
+    my $kernname;
+    foreach (
+        "image",   # s390x
+        "vmlinux", # ppc*
+        "Image",   # aarch64
+        "uImage",  # armv* (legacy)
+        "zImage",  # armv*
+        "vmlinuz", # x86*
+        ) {
```
schaefi: Hmm, actually this should be handled properly in KIWIConfig.sh here:
https://github.com/openSUSE/kiwi/blob/master/modules/KIWIConfig.sh#L1548

The cases you have added here are handled there too, we were also able to build for arm in the past
Thus I'm confused

dirkmueller: ok, the KIWILinuxRC.sh change might be redundant then, I'll remove it. 

The main issue I ran into is that extractkernel was not copying out the kernel it used into the destination directory, so steps further down the line failed on that.

schaefi: yes and that issue should not be present if the code in KIWIConfig.sh around line 1548 would have matched it. You should also see a log message ''Failed to find a mapping kernel' in the kiwi log file in this case

dirkmueller: ah, I see, I wasn't calling suseStripKernel, so thats why. how about the following patch instead (untested)?

schaefi:
looks good to me

PR #521, opened 2015-11-25
- Update copyBootCode function:

ggardet:
Add support to check for /boot/dtb* since DTB files
	can be stored in /boot/dtb/ folder but also in
	/boot/dtb-<version>/ folder with /boot/dtb a symlink


schaefi:
Thanks for adapting this

PR #522, opened 2015-11-27
fix spec file parts for RHEL/CentOS 7

seife:
This is necessary to be at least somehow able to build a CentOS7 image in OBS.
I tested this in my private OBS instance (more stuff is necessary and tweaks not easily possible in the openSUSE OBS).

The changes basically are:
* provide / require "yum" instead of "zypper"
* require grub2 instead of grub (no grub in CentOS7 anymore)

Note that this has only be tested with centos7, even though some of the hunks might also affect centos6
Example:

    -%if 0%{?suse_version} < 1315
    +%if 0%{?suse_version} < 1315 && 0%{?rhel_version} < 700 && 0%{?centos_version} < 700

But this is actually a no-op, because on centos < 700, 0%suse_version < 1315 will also be true :-)

### DIFF
--- a/rpm/kiwi.spec
+++ b/rpm/kiwi.spec
@@ -168,17 +168,13 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 %define mysystems suse-SLES%{sles_version}
 %endif
 %endif
-# RHEL
-%if 0%{?rhel_version} == 600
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks much for fixing that part, looks good to me

PR #523, opened 2015-11-27
make .checksum.md5 contents reproducible

seife:
In order to have reproducible builds in OBS with unchanged sources, sort
the file list before feeding it to "md5sum". Without, rebuilds would
differ in that file as the file list order was random.
Also, get rid of "find | grep -v ... | grep -v ..." for ignoring
filenames and just use find's options to achieve that. Saves two fork()
and a few CPU cycles :-)


schaefi:
yeah, good patch, thanks man

PR #524, opened 2015-11-27
Centos grub mkimage fix

seife:
grub2-mkimage on CentOS/RHEL7 *needs* the "-p <prefix>" option or it
will abort with an error.
On openSUSE's grub2-mkimage, prefix defaults to "/boot/grub2".
Add "-p /boot/grub2" to all grub2-mkimage calls, so that they can work
with CentOS, too. This *should* not break things on SUSE base ;-)

So this should be harmless... but might be good to test with more different SUSE configurations.
I only tested a OEM ISO built with this config:

    <type checkprebuilt='false' boot='oemboot/suse-SLES12'
          fsnocheck='true' filesystem='ext3' bootloader='grub2'
          installiso='true' installpxe='true' installboot='install'
          installstick='true'
          kernelcmdline='net.ifnames=0 video=800x600'
          fsmountoptions='acl'
          firmware='uefi'
          image='oem'>

...and this booted / installed fine in a Virtualbox VM. But that's about it... ;-)

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -3894,6 +3894,7 @@ sub setupBootLoaderStages {
             my $core= "$tmpdir/EFI/BOOT/$fo_bin";
             my $core_opts;
             $core_opts = "-O $fo -o $core -c $earlyboot ";
+            $core_opts.= "-p /boot/grub2 ";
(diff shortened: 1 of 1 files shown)


schaefi:
I see "-p, --prefix=DIR     set prefix directory [default=/boot/grub2]" for all supported suse versions using grub2. So I agree this should be a safe change

Let's go for it

PR #528, opened 2015-12-09
Run dracut as background process for vmx images

schaefi:
The dracut process takes quite some time and blocks the boot
process. In cloud frameworks instance boot time matters and
with this patch the dracut process runs in parallel with the
rest of the boot process.


dirkmueller:
:+1:

PR #532, opened 2015-12-21
Don't change hybrid file system type

e9925248:


modules/KIWILinuxRC.sh
```diff
@@ -9712,6 +9700,35 @@ function reloadKernel {
...
+        e2fsck -p $device
+    elif [ "$FSTYPE" = "btrfs" ];then
+        btrfsck $device
+    elif [ "$FSTYPE" = "xfs" ];then
+        xfs_repair -n $device
+    else
+        FSTYPE=$FSTYPE_SAVE
+        # don't know how to resize this filesystem
+        Echo "Don't know how to resize ${FSTYPE}... skip it"
+        return
```
schaefi: the error message is misleading, guess this was just a copy & paste typo :)

schaefi:
The overall change looks good to me, thanks much

e9925248:
Checked the two texts. fat [= container with btrfs inside] is probably not handled correctly.

schaefi:
Thanks, merging

PR #533, opened 2015-12-22
Handle /boot/vc files for Raspberry Pi

afaerber:
openSUSE Tumbleweed raspberrypi-firmware[-branding-openSUSE] packages
have been updated to install files to /boot/vc rather than /boot.

Ensure that all files in /boot/vc find their way into the vboot partition
and on the boot partition prepare vc as potential mount point.

Signed-off-by: Andreas Färber <afaerber@suse.de>


schaefi:
Thanks looks great :+1:

PR #534, opened 2015-12-22
Allow xfs as hybrid RW file system

e9925248:
btrfs is still not fully ready [eg. it reliably breaks with overlay fs on Leap 42.1].  xfs is used by current openSuSE for data partitions [instead of ext3/4].


schaefi:
yes makes sense

e9925248:
KIWI images will need the following to use xfs:
        <drivers>
                <file name="fs/xfs/*"/>
        </drivers>
        <packages type="image">
                <package name="xfsprogs" bootinclude="true"/>

PR #536, opened 2015-12-26
fix message of missing command

cyberorg:
Live iso created on recent distributions gives out this error message when booting that it is missing systemd-detect-virt command.

Try one from here for example: http://download.opensuse.org/repositories/Education/images/iso/

This commit fixes the error.


schaefi:
Thanks, I will merge it in as soon as the review for exfat is done

PR #541, opened 2016-02-01
exclude kernel-default from vmx requirements

lnussel:
the vmxboot templates do allow selecting an alternate profile without
kernel. The dependency generator however always includes kernel-default
into requires which makes it impossible in practice to build an image in
OBS that uses kernel-default-base (which conflicts with kernel-default)

### DIFF
--- a/rpm/kiwi.spec
+++ b/rpm/kiwi.spec
@@ -429,7 +429,7 @@ Requires:       genisoimage
 Requires:       cdrkit-cdrtools-compat
 %endif
 Requires:       kiwi-desc-vmxboot = %{version}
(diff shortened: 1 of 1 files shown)


adrianschroeter:
So this would break the standard builds, because kernel-default is not available anymore?

lnussel:
as long as the kiwi description that is built specifies a kernel that should be fine. Only for the situation where you build a vmx image without a kernel but expect kernel and initrd from kiwi to be there this would break IMO

lnussel:
could we give it a try at least?

lnussel:
as agreed with @adrianschroeter I opened https://github.com/openSUSE/open-build-service/issues/1528 and will close this one.

PR #543, opened 2016-02-04
output correct sha256 file

lnussel:
whether or not the argument to $sha->addfile is absolute doesn't
actually matter as $sha->hexdigest never outputs any file name. We
have to add the file name ourselves.

Reason for the confusion is code in OBS's build script that under
certain conditions computes the sha256sum file itself. So this code here
was never used where it actually matters.

This reverts commit 97b92d0e0a58cdd237b109efbf843b6057172164.


schaefi:
ok

PR #544, opened 2016-02-04
always sign bundled result

lnussel:
no matter what the output format is we want OBS to add a gpg
signature on the result. Therefore always also produce a .sha256
sums file for that purpose.


schaefi:
Thanks looks great

PR #546, opened 2016-02-11
- Enable pvops builds for EC2

rjschwei:
Additional changes to KIWILinuxRC.sh are needed, detailed in an e-mail.

The symptom is as follows:

[    1.135490] Waiting for device LABEL to settle...

This stems from the fact that we now generate "root=LABEL=ROOT" in grub.cfg and the code assumes that whatever follows "root=" is a device or partition. This assumption was added to the code way back when we sorted out how to get EC2 images to behave nicely. This has to change with the emergence of the pvops kernel. I tried:

if [ ( ! -z "$root" ) -a ( "${root#*$devIndicator}" ) ];then

line 4720 in LinuxRC.sh but that results in no root device being found, so we need something else there.

Anyway, I realize this is incomplete but it gets us part of the way there.


schaefi:
yes thanks, looks good to me. I'll merge this one and will come up with a patch for the different root variable layout

PR #548, opened 2016-02-17
Do not activate dmraid partitions

rpasche:
This only activates the device itself, not the partitions, that may also exist. If partitions exist, the UDEV rules will create the corresponding "_partX" symlinks for each partition within "/dev/mapper, which are totally sufficient.


schaefi:
Good catch, thanks much

PR #549, opened 2016-02-20
Remove flush before resize filesystem

dirkmueller:
According to the manpage, -F is flushing the fileystem buffer
caches. which is only really useful for doing resize2fs time trials.

With current Tumbleweed installed on MMC, -F triggers a Inappropriate ioctl for device
while trying to flush error, and then fails to resize.


schaefi:
whatever works to resize the thing is fine with me, I remember we put that option in because it was considered useful for sd cards. However if that triggers an error now it should be changed

Thanks for the fix

PR #550, opened 2016-02-20
Remove grub2 from the armv6/v7 initrd

dirkmueller:
Grub2 adds 35MB to the initrd for no reason on armv6/v7, as
those images use u-boot scripts for booting.

### DIFF
--- a/system/boot/armv7l/oemboot/suse-SLES12/config.xml
+++ b/system/boot/armv7l/oemboot/suse-SLES12/config.xml
@@ -105,7 +105,6 @@
         <package name="fribidi"/>
         <package name="genisoimage"/>
         <package name="gettext-runtime"/>
-        <package name="grub2"/>
         <package name="grub2-arm64-efi" arch="aarch64"/>
         <package name="hwinfo"/>
         <package name="iputils"/>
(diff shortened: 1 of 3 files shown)


schaefi:
make sense. I also saw we are adding grub2-arm64-efi for the machines with an EFI interface. Does this package not pull in grub2 ?

matwey:
Also, could you please explain why is bootloader needed in oemboot initrd? Isn't it possible to create new initrd and update bootloader config after real root is found, resized and mounted? Then we could just run applications from / instead of loading them into memory.

dirkmueller:
@schaefi grub2-arm64-efi depends on grub2, so the change is a noop for aarch64

schaefi:
>    [1]@schaefi grub2-arm64-efi depends on grub2, so the change is a noop
>    for aarch64

ok that's what I thought too.

schaefi:
>    Also, could you please explain why is bootloader needed in oemboot
>    initrd?

when kiwi is required to create bootloader image e.g for efi it
takes the bootloader modules from the initrd and not from the system
image. It has turned out this was a bad design decision and is
fixed in the next generation code base (kiwi v8). I'm afraid I won't
refactor this in the current version of kiwi to avoid regressions

> Isn't it possible to create new initrd and update bootloader
> config after real root is found, resized and mounted? Then we could
> just run applications from / instead of loading them into memory.

I'm not sure if I understood it correctly, but we do that on first
boot of the appliance. the bootloader setup is rewritten to match
the now real geometry and device node name(s) and the kiwi initrd
is replaced by the result of dracut

If your next question would be why we can't use the dracut initrd
from the beginning on, this will also be addressed in the v8
code stream where we can provide a BootImageX class for the different
use cases. So far I have only implemented BootImageKiwi, but
have an issue open for BootImageDracut :-)

PR #551, opened 2016-02-21
Fix filesystem check on boot

dirkmueller:
In the refactoring fc363cc06, the parameters of checkFileSystem got
changed to expect the device name, but one of the invocations didn't
pass down the device name due to a typo.


schaefi:
Thanks much for finding this

PR #552, opened 2016-02-23
run exec >/dev/null after screen clears in quiet mode

cyberorg:
If exec >/dev/null is run before setterm clears the screen, then "Failed to find cpu0 device node" is displayed even if quiet is set. This commit fixes that.


schaefi:
all right :+1:

PR #556, opened 2016-03-04
modules: KIWIContainerBuilder: preserve xattrs

### LINKED ISSUE #555: Extended Attributes not preserved when creating Docker containers
The final `tar` archive created by `kiwi` doesn't preserve the xattrs (specifically ones like set-capabilities which are quite important for OS filesystem images).

This is the root cause of openSUSE/docker-containers-build#8.

cyphar:
tar doesn't preserve extended attributes by default, causing Docker
images to not have any correct set-capabilities bits set on binaries
such as ping. Fix this by adding the --xattrs flag to the tar command
run to generate the root filesystem image.

Fixes #555.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ before_install:
     - sudo apt-get update -qq
     - sudo apt-get install -y xsltproc
     - sudo apt-get install -y syslinux
+    - echo "deb http://archive.ubuntu.com/ubuntu trusty main" | sudo tee -a /etc/apt/sources.list
+    - sudo apt-get update
+    - sudo apt-get install -y -t trusty tar
 
 install:
     - cpanm -f -n $(cat .perlmodules | tr "\n" " ")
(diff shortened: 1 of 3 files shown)


cyphar:
The reason the tests are failing is because the tests are running on Ubuntu. Ubuntu doesn't have `tar>=1.27`, which means that `--xattr` support doesn't exist then. This is not the case on openSUSE or SLES. However, there simply *is no other way* to add the extended attributes to the tar archive (aside from re-implementing it in Go, whose standard library has a decent `archive/tar` implementation).

If we ran the testing inside Docker, we could solve this versioning issue.
Another option is to use `star`, which has had support for xattrs since `2003`. I doubt that it produces tar archives that can be fully understood by [tar-split](https://github.com/vbatts/tar-split), which is what Docker uses to deal with `tar` archives.

schaefi:
Thanks for the patch, looks all good to me

PR #558, opened 2016-03-09
Execute haveged in udevStart

rsassu:
This ensure there is at least a source of entropy required for reading
from /dev/random. Without it VMware virtual machines hang at boot
(bsc#964204).

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -750,6 +750,7 @@ function udevStart {
     # start the udev daemon.
     # ----
     local IFS=$IFS_ORIG
+    local fipsEnabled=0
     #======================================
     # Check time according to build day
     #--------------------------------------
@@ -802,11 +803,17 @@ function udevStart {
             enablePlymouth=0
             break
             ;;
+        case "$o" in
+            fips=1*)
+            fipsEnabled=1
+            ;;
         esac
     done
     if [ $enablePlymouth -eq 1 ]; then
         startPlymouth
     fi
+    # start haveged
+    startHaveged
 }
 #======================================
 # moduleLoadBeforeUdev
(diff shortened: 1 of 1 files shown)


modules/KIWILinuxRC.sh
```diff
@@ -750,6 +750,7 @@ function udevStart {
     # start the udev daemon.
     # ----
     local IFS=$IFS_ORIG
+    local fipsEnabled=0
```
schaefi: I would be more in favour of enableFips=0, it would be more in line with e.g enablePlymouth

sorry for nitpicking

modules/KIWILinuxRC.sh
```diff
@@ -802,11 +803,17 @@ function udevStart {
...
+            fips=1*)
+            fipsEnabled=1
+            ;;
         esac
     done
     if [ $enablePlymouth -eq 1 ]; then
         startPlymouth
     fi
+    # start haveged
+    startHaveged
```
schaefi: iirc the plan is to start haveged if fips is enabled, correct ? if so I think the following is more clear

```bash
if [ $fipsEnabled -eq 1 ];then
    startHaveged
fi
```

or am I misunderstanding ?

modules/KIWILinuxRC.sh
```diff
@@ -881,6 +888,20 @@ function startPlymouth {
...
+       /usr/sbin/haveged
+       return
+    fi
+
+    # a source of entropy is required for VMware virtual machines
+    if [ $fipsEnabled -eq 1 ]; then
+       echo "Warning: haveged is missing. The boot may hang if /dev/random is being read."
+    fi
+}
+#======================================
```
schaefi: Given the function is called only in case of fips enabled it could be imho written like the following

```bash
function startHaveged {
    if lookup haveged &>/dev/null; then
        if ! haveged; then
            systemException \
                "Failed to start haveged required for fips" \
            "reboot"
        fi
    else
        systemException \
            "haveged is missing but required for fips" \
        "reboot"
    fi
}
```

Thoughts ?

modules/KIWILinuxRC.sh
rsassu: Well, haveged is used also without fips mode enabled. But for me, the change is fine.

rsassu:
I opened #559. I'm closing this pull request.

PR #559, opened 2016-03-09
Execute haveged in udevStart

rsassu:
This ensure there is at least a source of entropy for /dev/random when
the fips mode is enabled. Without it, VMware virtual machines hang at boot
(bsc#964204).

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -750,6 +750,7 @@ function udevStart {
     # start the udev daemon.
     # ----
     local IFS=$IFS_ORIG
+    local enableFips=0
     #======================================
     # Check time according to build day
     #--------------------------------------
@@ -802,10 +803,17 @@ function udevStart {
             enablePlymouth=0
             break
             ;;
+        case "$o" in
+            fips=1*)
+            enableFips=1
+            ;;
         esac
     done
     if [ $enablePlymouth -eq 1 ]; then
         startPlymouth
+    fi
+    if [ $enableFips -eq 1 ];then
+        startHaveged
     fi
 }
 #======================================
@@ -878,6 +886,22 @@ function startPlymouth {
         plymouth show-splash &>/dev/null
         # reset tty after plymouth messed with it
         consoleInit
+    fi
+}
+#======================================
+# startHaveged
+#--------------------------------------
+function startHaveged {
+    if ! lookup haveged &>/dev/null; then
+        systemException \
+            "haveged is missing but required for fips" \
+        "reboot"
+    fi
+
+    if ! haveged; then
+        systemException \
+            "Failed to start haveged required for fips" \
+        "reboot"
     fi
 }
 #======================================


rjschwei:
IMHO it is the users responsibility to start haveged. This is done in config.sh and we do this for all images we build for openSUSE and SLES.

Additionally haveged is a service and we should refrain from executing the binary directly. If it is decided that kiwi should take this out of the hands of the user then the code should enable and start the service. The proposed implementation only runs haveged on first boot. After the initrd is regenerated the kiwi specific code is no longer executed and you'd end up with the same problem, i.e. e non booting system because there is not enough entropy, this of course depends on the key generation requirements.

schaefi:
I was under the impression that haveged is required while we are still operating in the initrd.

If this is not the case I agree to what Robert said

rjschwei:
haveged generates entropy which is needed to generate keys, at startup the keys that get generated are the ssh host keys. At the time the ssh host keys get generated we are well passed the initrd. There is no process, to the best of my knowledge that requires entropy in the initrd.

schaefi:
all right, thanks for clarifying that. So I wonder why this one came up then, Roberto ?

rsassu:
Entropy is necessary also for other programs reading from /dev/random (e.g. ps, logger in SLES 12).

rsassu:
This does not apply for util-linux-2.25-30.1.x86_64 (installed in SLES 12 SP1).

rsassu:
Well, it seems that /dev/random is read only when fips=1 has been specified in the kernel command line (regardless of the util-linux version).

rsassu:
@schaefi, given my last comment, will you accept the patch?

schaefi:
The patch is good however there is still the following which confuses me:

> There is no process, to the best of my knowledge that requires entropy in the initrd.

So is this really required in the initrd ? I don't see that haveged is started in a dracut created initrd.
So which process in the kiwi initrd requires that we start haveged ? If there is no such process the service could and imho should be started by the normal init system (systemd or sysvinit)

rsassu:
The code to start haveged in dracut is (module: 98haveged):
---
ln_r "$systemdsystemunitdir/haveged.service" "$systemdsystemunitdir/systemd-journald.service.wants/haveged.service"
---

This behavior seems to happen only if /proc/sys/crypto/fips_enabled is 1.

rsassu:
Ok, I probably used characters that changed the message format.

schaefi:
iirc the dracut module just activates haveged to be started by systemd but does not require it when the initrd is active. What we do with the patch is actively starting haveged and I still do not understand which provess while we are inside of the initrd requires the presence of haveged ?

rsassu:
Haveged is executed from the initial ram disk:


[  OK  ] Reached target Initrd File Systems.
[  OK  ] Reached target Initrd Default Target.
pre-pivot

Generating "/run/initramfs/rdsosreport.txt"


Entering emergency mode. Exit the shell to continue.
Type "journalctl" to view system logs.
You might want to save "/run/initramfs/rdsosreport.txt" to a USB stick or /boot
after mounting them and attach it to a bug report.


pre-pivot:/# mount -t proc none /sysroot/proc/
pre-pivot:/# chroot /sysroot/
sh-4.2# ps aux|grep haveged
root       113  1.2  0.3  12036  3808 ?        Ss   09:57   0:00 /usr/sbin/haveged -w 1024 -v 0 -F
root       416  0.0  0.0  10496   928 ?        R+   09:58   0:00 grep haveged

schaefi:
ok thanks for checking all this

PR #560, opened 2016-03-18
bootImage: don't copy initial ram disk content to /run/initramfs

rsassu:
We should avoid copying the initial ram disk content to a tmpfs filesystem,
especially on low memory systems.


schaefi:
Thanks, is this not related to bnc#963276 ? if so it would be nice to see that in the commit message
It makes it easier to cherry pick into SLE11/SLE12 branches

PR #561, opened 2016-03-21
have tar preserve all xattrs

brauner:
By default tar will only preserve a subset of the capabilities with --xattrs.
Adding --xattrs-include=* should preserve all capabilities (e.g. user.*,
security.capability etc.).

Signed-off-by: Christian Brauner <christian.brauner@mailbox.org>


schaefi:
Thanks much

PR #562, opened 2016-03-22
bootImage: don't copy initial ram disk content to /run/initramfs

rsassu:
We should avoid copying the initial ram disk content to a tmpfs filesystem,
especially on low memory systems (bsc#963276).


rsassu:
I fixed the commit message.

schaefi:
Thanks, looks great now

PR #563, opened 2016-03-22
Bsc963676 slepos11 sp3

rsassu:
Backported patches:
- 98c2d64 - Prefer switch_root over pivot_root
- 5ca2108 bootImage: don't copy initial ram disk content to /run/initramfs


schaefi:
Thanks, great work :+1:

PR #566, opened 2016-04-04
plugin for new createrepo_c in kiwi

M0ses:
@schaefi : We are planning to move to createrepo_c in obs, therefore the new plugin.


schaefi:
nice :+1:

PR #567, opened 2016-04-14
Build kiwi-desc-isoboot on other architectures

k0da:
geniosimage is not x86 specific, syslinux dependency is guarded there.


k0da:
Hmm.. no incomplete

schaefi:
ok please re-open when you are complete, thanks

PR #572, opened 2016-05-03
new sle12-sp2 media style

adrianschroeter:
default patterns are now provides and not recommends

Code is already packaged, just merge :)


schaefi:
ok

PR #573, opened 2016-05-05
Add aarch64_ilp32 support

dirkmueller:
This is the -32bit variant of aarch64. Right now only needed for
install medias.


schaefi:
@dirkmueller it seems like #575 is nearly the same, which one is correct ?

dirkmueller:
@schaefi well, mine is 4 days earlier and fixes two places instead of one ;-) Other than that I don't care really, we just need it to unbreak Factory..

schaefi:
ok :smile:

PR #575, opened 2016-05-09
Add aarch64_ilp32 to the list of known architectures

andreas-schwab:


schaefi:
It seems like #573 does nearly the same, which one is correct ?

andreas-schwab:
I don't know, mine is tested.

PR #579, opened 2016-05-24
tag debug medias in rpm-md meta data (bnc#980871)

adrianschroeter:
package for SLE already submitted


adrianschroeter:
opensuse package is on the way as well now: request 397826

PR #583, opened 2016-05-30
fix hash key handling for perl 5.24 (boo#981080)

adrianschroeter:


schaefi:
Thanks much, looks good to me :+1:

PR #584, opened 2016-06-17
Fix RHEL6 bootloader install

k0da:
RHEL expects /etc/grub.conf to be a symlink to /boot/grub/grub.conf,
lets fix a condition where we enter that specific path.


schaefi:
:+1: 

Thanks

PR #585, opened 2016-06-17
Don't run updateMTAB on RHEL6 systems

k0da:
RHEL6 doesn't have /etc/mtab as a link. Let's keep original behaviour


schaefi:
:+1:

PR #587, opened 2016-07-09
prefer newer overlayfs

cyberorg:
Prefer newer overlayfs otherwise mount succeeds in compat mode but live image fail to boot/exhibit weird symptoms.

Affects 42.2 and tumbleweed live


schaefi:
ok but the comment needs to change as well, now it's misleeding. Could you adapt that part too ?

Thanks

schaefi:
all right let's go for it, thanks :+1:

PR #589, opened 2016-07-18
Rhel netboot

k0da:


modules/KIWIXMLPXEDeployData.pm
```diff
@@ -150,6 +150,8 @@ sub createPartition {
     my $this     = shift;
     my $partInfo = shift;
     my $kiwi = $this->{kiwi};
+    my $firmware = $this->{firmware};
+    $MAX_PART_ID = '256' if $firmware =~ /efi/;
```
schaefi: where is this used ? I see it only declared

k0da: @schaefi https://github.com/k0da/kiwi/blob/79e22d761db864730a0ad7749cabcbec17d488d7/modules/KIWIXMLPXEDeployData.pm#L185

schaefi:
all right :+1:

PR #591, opened 2016-07-27
Instsource tumbleweed

DimStar77:
This is intentionally done in two commits:

1) realy duplicate 13,2 to Tumbleweed - separating this out gives the freedom to not break other setups relying on this (kiwi-instsource-plugins-openSUSE is so far also used in Leap 42.2)
2) Update KIWIMini to add an additional file into initrd: instruct linuxrc to use kexec=1 (that is: download initrd from the network as well and respawn the installer)

note: I had close to no chance yet to test this... once we have a kiwi-instsoource-plugins=openSUSE-Tumbleweed in Tumbleweed I can adjust the _product definitions to use the new templates


schaefi:
looks good to me

PR #593, opened 2016-08-22
waitForStorageDevice patch

davidcassany:
This solves the issue bnc#992992. Now waitForStorageDevice method
has set a timeout of 4 seconds instead of 60 if the device is a usb.
GetDeviceTransportType method has been added in order to determine
device type (usb, sata, etc.)

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -7395,6 +7395,21 @@ function killShell {
     fi
 }
 #======================================
+# getDeviceTransportType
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks David, small typo in the commit message, it should be bnc# not bcn#. The buildservice tracker will most probably not track this correctly. Could you "git rebase -i master" your  change, and "r = reword" the commit message. After that you can "git push force" to update here

Thanks

schaefi:
:+1:

PR #596, opened 2016-09-18
Use more reliable tool to detect link status, fall back to ip if it is not available

cyberorg:
On some clients 'ip link ls' shows 'state UNKNOWN' using ifplugstatus seems to be more reliable so prefer that if it is available, ifplugd which provides ifplugstatus command is only added to tumbleweed config.xml at the moment, may be it should be added to other distrbiutions  netboot config.xml as well.

This is proper version of https://github.com/openSUSE/kiwi/pull/595
boot.kiwi log broken client: https://github.com/openSUSE/kiwi/files/478114/boot.txt
boot.kiwi log with this fix applied:
[boot.txt](https://github.com/openSUSE/kiwi/files/478640/boot.txt)

### DIFF
--- a/metadata/KIWIConfig.xml
+++ b/metadata/KIWIConfig.xml
@@ -131,6 +131,7 @@
         <file name="init"/>
         <file name="insmod"/>
         <file name="id"/>
+        <file name="ifplugstatus"/>
         <file name="iucv_configure"/>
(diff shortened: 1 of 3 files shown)


schaefi:
Thanks looks good to me

PR #597, opened 2016-09-22
Custom device timeout variable added

davidcassany:
This commit patches the issue bnc#992992. Now DEVICE_TIMEOUT
variable can be used in order to set the device waiting timeout
in watForStorageDevice function. Must be a numeric value expressed
in seconds.


schaefi:
:+1:

PR #598, opened 2016-10-01
do not use ramdisk for rw, using tmpfs is more efficient

cyberorg:
Defaulting to tmpfs has few advantages:
* RAM does not have to be reserved for ramdisk allowing applications to use RAM not use by the system.
* mkfs run is not needed on /dev/ram1


schaefi:
looks good to me, just to be sure, you have tested this change in an image build right ?

cyberorg:
Yes, I've tested this.

PR #599, opened 2016-10-03
Added support for mkisofs in addition to genisoimage

davidcassany:
Support for live iso images using mkisofs or genisoimage. Kiwi v7 always looks for a genisoimage executable in PATH, however now genisoimage can be a symlink to mkisofs. 

This PR is related to the issue bnc#997443.

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -925,7 +925,7 @@ sub setupInstallCD {
     my $attr = "-R -J -f -pad -joliet-long";
     if (-s $system >= 4294967296) {
         # install image is bigger than 4g, needs extra iso options
-        $attr .= " -allow-limited-size -udf -hfs -iso-level 3";
+        $attr .= " -udf -hfs -iso-level 3";
     }
     $attr .= " -V \"$volid\"";
     $attr .= " -A \"$this->{mbrid}\"";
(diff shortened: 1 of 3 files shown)


schaefi:
Thanks David, it looks good to me. Did you create a test image with that changes ? I'm just asking because we will be submitting this as maintenance update to SP2 and we have no integration testing like we have for v8. Thanks

davidcassany:
@schaefi I built a Leap 42.1 Jeos (from kiwi-decriptions) including a 4Gb file and tested the resulting iso on USB in my lap and also the iso file in a libvirt machine as CD-ROM source. I also verified the filesystem is UDF. I did not do any further testing than that (like using different firmware combinations an so on).

Also I did not want to remove cdrkit-crtools-compat package requirement from spec without discussing it (I guess that has to be ported only to specific versions). I think the way to go is to remove that package requirement and keep genisoimage requirement as is (kiwi v8 still has genisoimage requirement and this way we don't to add a checker for mkisofs command in kiwi v7). Also we can make that sort of changes in a separate PR. Between, in my host I don not have installed that compat package, I ignored that dependency in kiwi v7 packages and I am not experimenting any issue as genisoimage is still a requirement.

schaefi:
Thanks sounds all good :+1:

PR #600, opened 2016-10-10
Remove cdrkit-cdrtools-compat from requirements

davidcassany:
cdrkit-cdrtools-compat requirement is no longer needed as we aim to be transparent to mkisofs and genisoimage commands.

This PR is related to bnc#997443.


schaefi:
:+1:

PR #601, opened 2016-10-11
Solving some broken dialogs

davidcassany:
HideSplash method used to hide the splash only if there was an active console, however the dialogs appear broken if the splash isn't hidden and in some cases is hard to check if there is an active consoles (i.e. after umount /sys or mount --move /dev /mnt/dev).

This commit fixes bnc#997085

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -115,21 +115,17 @@ failsafe="$failsafe nomodeset x11failsafe"
 #--------------------------------------
 function hideSplash {
     # /.../
-    # Only in case of one active console kiwi hides
-    # the splash screen for it to allow interactive dialog
-    # sessions on this console. In any other case the user
-    # can control a custom behavior using the handleSplash
+    # Hides the splash screen for to allow interactive 
+    # dialog sessions on this console. Also, the user can 
+    # control a custom behavior using the handleSplash
     # hook called at the end of this function
     # ----
     local IFS=$IFS_ORIG
-    local console_count=$(activeConsoles)
-    if [ $console_count -eq 1 ];then
-        test -e /proc/splash && echo verbose > /proc/splash
-        if lookup plymouthd &>/dev/null;then
-            plymouth hide-splash
-            # reset tty after plymouth messed with it
-            consoleInit
-        fi
+    test -e /proc/splash && echo verbose > /proc/splash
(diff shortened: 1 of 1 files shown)


schaefi:
Looks good to me, we need this in the v8 code base too. The patch should apply without conflicts to kiwi/boot/functions.sh. Do you want me to port it over once this is merged or do you care for it too ?

Thanks

schaefi:
:+1:

davidcassany:
I will port it to kiwi v8, it will be easy, I have already located the code.  thanks

schaefi:
Thanks that's great

schaefi:
Another point to discuss. The commit message says this commit tries to fix it... that doesn't sound very confident :smile: So do we know if it fixes the broken dialog or do we just guess ? If we know it fixes a certain condition we should explain that condition and raise our hands for other scenarios.

Thoughts ?

davidcassany:
Yes, you are right. I know it solves some some issues, concretely, it solves the dialogs that appear if oem-reboot-interactive or oem-shutdown-interactive flags are set. However I have not tested the progress dialogs that may appear in a netboot (I guess due to fetchFile method), I tried to set a virtual server/client env but I did not succeed up to now. I suspect in that concrete case it might happen that the hideSplash method is never called, thus the splash keeps being visible and if the user presses 'ESC' key then a broken dialog with a progress bar should appear. In that concrete case it is not that critic since the progress bar wouldn't be demanding any user interaction, so the boot would just run fine, but with an ungly dialog trying to show a download progress. All that said, this is just a guess, as I could not test it yet.

schaefi:
That's fair enough, so if you don't mind remove the "tries" from the commit message and say that it fixes the broken dialog scenarios reported in bnc#997085 because that's the truth. If the QA testing will find that the bug is only partially fixed we can reference a follow up fix on the same bug. But that we tried our best and failed will not highlight the work and effort you spent on it :)

davidcassany:
Ok done. However I managed to reproduce the bug with a pxe image, as I suspected hideSplash is not called in the fetchFile method. I am about to include another commit to fix it. Now testing it, I am pretty sure that adding hideSplash call will be suficient.

schaefi:
cool, I'll wait with the merge ;)

davidcassany:
Ok, now I definitely tested it in all the scenarios reported in bnc#997085. Those broken dialogs are fixed now.

PR #602, opened 2016-10-11
Fixed creation of etc/crypttab

schaefi:
The method setupCryptTab creates etc/crypttab from the given
storage device. However in a raid setup the correct /dev/mdX
device needs to be used and not one of the storage devices in
the raid array. Fixes bnc#1003595


nadvornik:
LGTM

PR #605, opened 2016-10-19
Fixed a rounding error in partedMBToCylinder method

ajchace:
bc truncates number of cylinders to zero decimal places, which results in a
  partition that is slightly smaller than the requested size. Add one cylinder
  to compensate.


schaefi:
yep patch is great

@davidcassany  would be perfect if you can port it over to v8

Thanks

PR #606, opened 2016-10-25
No longer write a %changelog section in the spec

schaefi:
For some reason the packaging concept no longer needs a changelog
section in the spec file. Most likely the build service scripts
now generates those information from the .changes file which lead
to duplicate changelog information in the package


schaefi:
@lnussel Hi, would you mind to give this one a short review. It seems we don't want the spec file to contain a %changelog section anymore. At least Jan Engelhard requested to drop that because the changelog would then be duplicated

Thanks

lnussel:
you only need an empty %changelog tag at the end of the spec file. The build script handles the rest.

schaefi:
ok, that should be the case now

PR #607, opened 2016-10-27
Assemble mdraid

schaefi:
Fixup assembling of mdraid array
    
when udev discovers an mdraid array it partially starts the array.
That is interfering with the mdadm --assemble call by kiwi which
leads to a busy state and an array in inactive state. Therefore
the method should wait until the raid array really exists no
matter if the assembling is started by udev or kiwi's mdadm call.
    
In addition if the array got assembled but is incomplete because
devices are missing or the timout is fired, an additional call to
start any array that has been partially assembled is required.
    
pxeRaidAssemble will throw an exception if after this call no
md device with a size > 0 will show up after a timeout. This
Fixes bnc#1000742

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -10525,6 +10525,15 @@ function pxeRaidAssemble {
         fi
         IFS=$IFS_ORIG
         mdadm --assemble --run /dev/md$mdcount $devices
+        if ! waitForStorageDevice /dev/md$mdcount; then
+            # start any array that has been partially assembled
+            mdadm -IRs
+            if ! waitForStorageDevice /dev/md$mdcount; then
+                systemException \
+                    "Failed to assemble raid array, too many devices missing" \
+                "reboot"
+            fi
+        fi
         mdadm -Db /dev/md$mdcount >> $conf
         mdcount=$((mdcount + 1))
     done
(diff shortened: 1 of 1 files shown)


modules/KIWILinuxRC.sh
```diff
@@ -10525,6 +10525,15 @@ function pxeRaidAssemble {
         fi
         IFS=$IFS_ORIG
         mdadm --assemble --run /dev/md$mdcount $devices
```
nadvornik: It the code above still needed? On SLE12 it will just generate error messages.

schaefi: seems my comment by mail got lost

for sle12 you are correct, but this is the v7 code base and that is also used for sle11 builds. Thus I don't want to break it

schaefi:
>    @nadvornik commented on this pull request.
>      __________________________________________________________________
> 
>    In [1]modules/KIWILinuxRC.sh:
> > @@ -10525,6 +10525,15 @@ function pxeRaidAssemble {
>          fi
>          IFS=$IFS_ORIG
>          mdadm --assemble --run /dev/md$mdcount $devices
> 
>    It the code above still needed? On SLE12 it will just generate error
>    messages.

not needed for sle12 that's true but this v7 also supporting sle11
Thus I don't want to break it

PR #608, opened 2016-10-28
Put setupNetworkWicked in line with former dhcpcd

schaefi:
The way setupNetworkWicked implements the dhcp discovery was
incomplete. This Fixes bnc#1003091

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -5222,24 +5222,14 @@ function setupNetworkWicked {
     local nic_config
     local dhcp_info
     local wicked_dhcp4=/usr/lib/wicked/bin/wickedd-dhcp4
-    for try_iface in ${prefer_iface[*]}; do
-        # try DHCP_DISCOVER on all prefered interfaces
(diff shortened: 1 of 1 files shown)


modules/KIWILinuxRC.sh
```diff
@@ -5256,19 +5246,39 @@ function setupNetworkWicked {
...
-        if [ -s $dhcp_info ]; then
-            export PXE_IFACE=$try_iface
-            ip route add default dev $PXE_IFACE
-            if [ ! -z "$GATEWAYS" ];then
-                # Use first entry as primary gateway
-                local gw=$(echo $GATEWAYS | cut -f1 -d " ")
-                ip route change default via $gw dev $PXE_IFACE
+        if [ -s $dhcp_info ] && grep -q "^IPADDR=" $dhcp_info; then
+            echo '<request type="lease"/>' |\
+                wicked test dhcp4 $try_iface > $dhcp_info
```
mtomaschewski: There is a ```--request -``` argument missed. Without it, stdin is not used.

PR #610, opened 2016-11-03
UpdateOtherDeviceFstab method was not keeping the first argument

davidcassany:
This PR fixes bnc#1007765

updateOtherDeviceFstab is method responsible to include the fstab
lines needed to mount additional partitions (apart from swap and
root) defined in the description file. This method had an error
overwriting the variable $1 with the value of $2, which is always
empty, as the method it is never used with more than a single
parameter.


schaefi:
wow what a bad bug, shame on me. Thanks for finding this. Could you port it over to v8 ? I see the same bug there. Thanks much

davidcassany:
yes sure I am porting it to kiwi v8 right now

PR #613, opened 2016-11-07
ActivateImage links correction and read-only test before copying

### LINKED ISSUE #612: Wrong copy commands in isoboot scripts when using read-only filesystems, without unionfs
## Problem 

In kiwi if trying to build iso images without any specific configuration, assuming the default behavior, leads to an image that fails on boot. Using a minimalist type configuration like the following:
`<type boot="isoboot/suse-SLES12" image="iso"/>`
while booting it fails in `activateImage` method trying to copy killall5 file from `/sbin/killall5` to `/mnt/sbin/killall5`, however the later is not a valid path during the preinit phase. Moreover the filesystem is read-only so even the paths are not correct the `cp` statement would fail.

davidcassany:
This PR is related to bnc#1006834  and fixes #612. 

Loading the iso image into a read-only ramdisk caused some issues due to the activatImage method cp commands. With this commit the symlinks are consistent during the preinit phase and also before running the cp command it is tested we are not in a read-only filesystem.

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -8220,6 +8220,7 @@ function activateImage {
     local cdDir=/livecd
     if [ -d $cdDir ];then
         mkdir -p $prefix/$cdDir && mount --move /$cdDir $prefix/$cdDir
+        rm -r $cdDir && ln -s $prefix/$cdDir $cdDir
         if [ -d /cow ];then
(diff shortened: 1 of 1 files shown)


schaefi:
looks good to me, we need this for v8 too. Could you create a pull request there too ? Thanks

davidcassany:
@schaefi sure I already did it in my local repos I'll push it now

PR #615, opened 2016-11-08
Testing isoinfo binary correctly

### LINKED ISSUE #614: Iso images builds are failing
## Problem

Images are failing in findAndCopyMagicBlock method, the magic block is not found. It has been experienced in kiwi-instsource in test DVDs from Factory in OBS.

davidcassany:
In method 'isols()' isoinfo command line tool is used, however the
binary was opened before testing if it existed. Apparently the open
perl function was not returning an error if the file did not exist.
So in the case that isoinfo was not located in /usr/bin/isoinfo the
open command was not failing, hence never trying to open the
fallback location /usr/lib/genisoimage/isoinfo.

This commit fixes #614.

### DIFF
--- a/modules/KIWIIsoLinux.pm
+++ b/modules/KIWIIsoLinux.pm
@@ -999,10 +999,15 @@ sub isols {
     my $isoinfo_opts = " -R -l -i $iso 2>/dev/null |";
     my $files;
     local $_;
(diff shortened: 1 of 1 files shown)


schaefi:
stupid perl, and again it strikes back that there are no tests for this code base :(

Thanks much for catching this one

PR #616, opened 2016-11-10
Add reading of kernel boot params from grub.cfg

schaefi:
The method getKernelBootParameters has been extended to be
able to read the kernel boot params also from boot/grub2/grub.cfg
This fixes bnc#1009032

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -3861,10 +3861,11 @@ function getKernelBootParameters {
         $prefix/boot/grub/menu.lst
         $prefix/etc/lilo.conf
         $prefix/etc/zipl.conf
+        $prefix/boot/grub2/grub.cfg
     "
     for c in $files;do
         if [ -f $c ];then
-            params=$(cat $c | grep 'root=' | head -n1)
+            params=$(cat $c | grep -v 'set root' | grep 'root=' | head -n1)
             break
         fi
     done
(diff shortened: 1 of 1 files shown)


modules/KIWILinuxRC.sh
```diff
@@ -3873,6 +3874,7 @@ function getKernelBootParameters {
     params=$(echo $params | sed -e 's@^module@@')
     params=$(echo $params | sed -e 's@^kernel@@')
     params=$(echo $params | sed -e 's@^parameters=@@')
+    params=$(echo $params | sed -e 's@^linux .* root@root@')
```
nadvornik: IMHO something like s@^linux [^ ]*@@ would be safer

nadvornik:
It works for my case, however I would not rely on root being the first parameter, as I mentioned above.

Also, it should better handle the situation when no bootloader config is found - I'd suggest to print a warning and maybe fallback to something like KIWI_KERNEL_OPTIONS.

schaefi:
actually I think the entire getKernelBootParameters() method is crap and full of possible error conditions.
How about changing this and always use the current /proc/cmdline information ? Would there be any information missing. What do you think ?

nadvornik:
There are some differences, for example resume= parameter is missing.

schaefi:
I have refactored the code to let each of the bootloader config methods write a metadata file. This file serves as a reference. That way we can avoid the scary parsing of the bootloader config files

Thoughts ?

PR #619, opened 2016-11-15
Lazy umount for the chrooted environment

davidcassany:
With this change any mount point that fails to normally umount,
is marked to lazy umount.

This PR is related with the issue bnc#1008898


davidcassany:
@schaefi as you will see, this is a bit different from what I explained you before about --bind mount. I changed the mount command as it is in Kiwi v8 and I saw the issue was still there, when I tried it last Friday maybe I had already changed the lazy umount also. Anyway, now it always tries to lazy umount in case a normal mount is failing. This is a similar behavior to what is implemented in root_bind in kiwi v8, which always performs a lazy umount of the chrooted root_bind, it always calls the umount_lazy() method.

schaefi:
ok that sounds all good to me, thanks :+1:

PR #620, opened 2016-11-24
Fixed expansion of msdos disks

schaefi:
If the target disk size is beyond 2TB it can't be expanded to
the full size if the msdos partition table layout is in use.
Because of this the disk expansion will be limited to the
allowed maximum for the msdos partition table type which is
at 2TB. This fixes bnc#1010966

### DIFF
--- a/system/boot/armv7l/oemboot/suse-repart
+++ b/system/boot/armv7l/oemboot/suse-repart
@@ -934,7 +934,16 @@ function getFreeDiskBytes {
     if [ ! -b "$disk" ];then
         echo 0 ; return
     fi
+    local pt_table_type=$(blkid -s PTTYPE -o value $disk)
     local disk_Bytes=$(blockdev --getsize64 $disk)
+    # max msdos table geometry is at 2TB - 512
+    # max sector count taken from parted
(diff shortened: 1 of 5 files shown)


system/boot/armv7l/oemboot/suse-repart
```diff
@@ -934,7 +934,16 @@ function getFreeDiskBytes {
     if [ ! -b "$disk" ];then
         echo 0 ; return
     fi
+    local pt_type=$(blkid -s PTTYPE -o value $disk)
     local disk_Bytes=$(blockdev --getsize64 $disk)
+    # max msdos table geometry is at 2TB - 512
+    # max sector count taken from parted
+    local max_dos_disk=$((4294967295 * 512))
+    if [ "$pt_table_type" = "dos" ] && [ "$disk_Bytes" -gt $max_dos_disk ];then
```
rjschwei: Where does pt_table_type get initialized? Is this transferred from some other script?

On line 937 pt_type is being set.

schaefi: This is why a review makes sense when stupid people like me create code

PR #622, opened 2016-11-28
Control volume variables order in .profile

### LINKED ISSUE #621: Wrong mount order when specifing nested mount points in LVM groups
If several partitions are defined for a _vmx_ image and they contain nested mount points, as in the `systemdisk` configuration shown below, `/var/log` get mounted before `/var`. Apparently this is due to the names, as the volume attributes are stored in a HASH where the key for each volume is the volume name. Then looping over that HASH results in an alphabetical order which ends up also setting the mount order.

```xml
<systemdisk name="vg0">
  <volume name="@root" size="4G"/>
  <volume name="LVhome" size="1G" mountpoint="home"/>
  <volume name="LVtmp" size="2G" mountpoint="tmp"/>
  <volume name="LVopt" size="1G" mountpoint="opt"/>
  <volume name="LVvar" size="2G" mountpoint="var"/>
  <volume name="LVlog" size="2G" mountpoint="var/log"/>
</systemdisk>
```

The user should not have to care about that, mount should happen in appropriate order.

davidcassany:
This PR modifies the order criteria of the .profile variables used by
the boot code during the first boot.

Fixes #621

### DIFF
--- a/modules/KIWIProfileFile.pm
+++ b/modules/KIWIProfileFile.pm
@@ -348,9 +348,34 @@ sub writeProfile {
             return;
         }
         binmode ($PROFILE, ":encoding(UTF-8)");
+
+        # kiwi_LVM_LV* varibles need to be included in .profiles with a
+        # correct order depending on the specified mount point if any.
+        # LVRoot is forced to be the first one and then depending on the
+        # depth level of the mount point. This prevents mounting nested
+        # mount points in wrong order during the first boot.
+        my %paths = ();
         foreach my $key (sort keys %profile) {
-            $kiwi -> loginfo ("[PROFILE]: $key=\"$profile{$key}\"\n");
-            print $PROFILE "$key=\"$profile{$key}\"\n";
+            if ($key !~ /^kiwi_LVM_.*$/) {
+                $kiwi -> loginfo ("[PROFILE]: $key=\"$profile{$key}\"\n");
+                print $PROFILE "$key=\"$profile{$key}\"\n";
+            } elsif ($key =~ /^kiwi_LVM_LVRoot$/) {
+                unshift @{$paths{0}}, $key;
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi could you please have a close look on this one. I am not sure the provided patch is the best way to fix this issue, I could not think up in any better solution without having to refactor some things. 

Also I am not very into Perl, so I maybe there are better ways to code the same algorithm, I tried to use a similar code to what is used in KIWIBoot.pm::setupBootDisk().

schaefi:
Thanks, I'll take a look. Could you check if Path.sort_by_hierarchy in the kiwi v8 code base works as expected regarding the volume example you used ? That would be nice to make sure we are not facing the same problem in v8. Thanks

davidcassany:
sure, testing Kiwi v8 was already in my TODO list for today

schaefi:
This is one of the changes which scares me, because it's hard to estimate if we are not opening a regression. I need a little bit more time for testing, sorry if this still sits for a while

davidcassany:
@schaefi, yes I know, I am also a bit concerned about that, so that I told you to have a close look. Don't worry if it stays long here, we always have the workaround to rename the volumes in a way the alphabetical order is good for our purposes.

Do you have in mind any testing I can help with, I tried this with some configs in centos and opensuse builds. Basically the systemdisk I posted in the issue with different volumes names and mountpoints (including the root slash and without including it).

schaefi:
I'm not aware of a better testing in this situation, I also started several combinations of <volume> setup and just run builds and boot tests

schaefi:
ok, with sweating hands I'll merge this, couldn't find a test case which broke it.

keep fingers crossed that this does not open a regression somewhere else

davidcassany:
I hope it doesn't, I'll keep my fingers crossed

PR #627, opened 2017-01-25
Validate package memtest86+ is present for Memory Test

davidcassany:
This commits adds a validation to verify memtest86+ package is
included in the kiwi description file before including the
'Memory Test' entry in the grub configuration. It applies
for ISO images.

This patch fixes bnc#1014150

### DIFF
--- a/modules/KIWIImage.pm
+++ b/modules/KIWIImage.pm
@@ -1138,6 +1138,22 @@ sub createImageLiveCD {
         $rootTarget = $cmdL->getConfigDir();
     }
     my $checkBase = $rootTarget."/".$baseSystem;
+
+    #==========================================
+    # Check Memory Test dependecies
+    #------------------------------------------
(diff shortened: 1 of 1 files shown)


schaefi:
yep fully agreed. On a short look the patch looks great. I just noticed the commit message does not contain the bugzilla bug id. Would be good to have it there because that ends up in the changelog and from there the info is mandatory for the suse package processing when we submit this to sles

Thanks

schaefi:
👍

davidcassany:
@schaefi, ok, thanks for the bugzilla id note, I thought it was enough to include it in the PR message. I'll update the commit message.

schaefi:
yep the kiwi changes file "kiwi.changes" in the package is created from the git changelog. The pull request message afaik is not part of the git log. The same applies to the v8 version. I have scripts which generated the changes file from the git log

just as a side information

PR #629, opened 2017-01-26
Fixup setupMachineID

### LINKED ISSUE #628: Interactive systemd-firstboot
## Problem description

Deleting `/etc/machine-id` and dbus id files triggers _systemd-firstboot_ service. Which is a one shot service that runs when `/etc/machine-id` is missing.

## Actual behaviour

During first boot there are some interactive settings, actually _systemd-firstboot_ asks for locales, time zone and some other settings.

## Steps to reproduce the behaviour

Run a Leap 42.1 vmx that has been build with the KIWI v7.04.24. I should also effect SLE and openSUSE versions though.

Side effect of 0c9fa69901738b17dd2db378cf0f873831449ed7

schaefi:
Cleaning up existing machine id files by deleting them
causes an interactive session to be started by systemd
This is something we don't want. As the consequences
of touching the machine id files seems to be too critical
the method has been turned into a hook caller. This
allows the user to make use of it on their own purpose
and by default doesn't mess with the machine id files
This Fixes #628

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -8105,19 +8105,11 @@ function setupConfigFiles {
 #--------------------------------------
 function setupMachineID {
     # /.../
-    # systemd-machine-id-setup initializes the machine ID
-    # in /etc/machine-id. The machine ID is defined to be a
-    # unique information. Thus it is required to initialize
-    # it on first boot of the image. In addition the same
-    # machine-id is configured to be used by dbus. In order
-    # to achieve this kiwi cleans up all existing machine
-    # id files which triggers the creation of a new set of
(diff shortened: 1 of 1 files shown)


schaefi:
This change also is in line with what was discussed here
https://trello.com/c/45cLtckf/419-suse-manager-3-0-appliance

rjschwei:
OK, fair enough. This puts the machine-id handling in the hands of the image builder. Those building images where the machine ID is basically immaterial get the behavior prior to the start of the episode of messing with the machine-id, and those that do care about the machine-id being duplicate can manage the behavior in the hook script. Users that do not use a kiwi generated initrd are on their own anyway.

Thanks, thumbsup

PR #631, opened 2017-02-03
Prevent quoted domain name

k0da:
We're getting domain name by parsion a lease file. Unfortunately in
lease file domain name is quoted, which breaks linux resolver.

This commit gets domainname unquoted


schaefi:
Thanks, looks good to me

PR #636, opened 2017-05-04
Fixed detection of Xen PV guest

schaefi:
The method isXen was used to check if the paravirtual kernel
kernel-xen is used and if so it influences the setup of the
bootloader. In newer versions there is no kernel-xen anymore
and its capabilities were moved into the pvops kernel provided
with the standard kernel-default package. Thus one kernel can
now serve for pv, hvm and dom0 images. Because of that there is
no other way than specifying the target Xen use case as part
of the image description. The information can be provided in
the doman setup of the machine configuratin with:

- domain="dom0"
- domain="domU"
- domain="domU-PV"

The new domU-PV value allows to specify that this image is going
to be used as a xen para virtual guest. The standard domU value
will setup the bootloader in a way that it is usable as "real"
bootloader in an hvmloader Xen guest. Fixes (bsc#1036198)

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -214,10 +214,6 @@ sub new {
     #------------------------------------------
     $this->{gdata} = $global -> getKiwiConfig();
     #==========================================
-    # check if Xen system is used
-    #------------------------------------------
-    ($xengz, $isxen) = $global -> isXen ($initrd);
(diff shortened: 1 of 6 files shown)


gmoro:
I validated the changes and it works ok for the cases I tested, can you please merge and release this change? Can I help in anyway?

PR #639, opened 2017-05-29
remove duplicated code

k0da:
IPADDR is assigned within dhclientImportInfo


schaefi:
Thanks looks good to me :+1:

PR #642, opened 2017-06-01
Fix baseSystemdCall method, fixes bsc#1041815

davidcassany:
This commit fixes the parameters passed to the systemd calls, with
the patch `baseService nfs-server off` can be executed without issues.


schaefi:
👍

PR #648, opened 2017-07-17
Do not include any suffix to the displayname value bsc#1029904

davidcassany:
If displayname attribute in the description file is used we should
not include any kind of suffix to that value. With this patch
the grub2 menu entry does not include the image type suffix if the
displayname option is used.

Fixes bsc#1029904

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -4296,11 +4296,14 @@ sub setupBootLoaderConfiguration {
     my $failsafe = 1;
     my $cmdline;
     my $title;
+    my $label_is_displayname = 0;
     #==========================================
     # set empty label if not defined
     #------------------------------------------
     if (! $label) {
         $label = "";
+    } elsif ($label eq $xml -> getImageDisplayName()) {
+        $label_is_displayname = 1;
     }
     #==========================================
     # Failsafe boot options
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi, sorry, this is not ready to merge, I have just realized that it only partially solves the issue. After the first boot and after applying the new dracut initrd image, the grub menu entry includes again the image type. I guess I need to update also something within the boot code.

davidcassany:
@schaefi, that should be fine now

schaefi:
I'm not sure why this is needed. afaik if we specify

```xml
<image ... displayname="foo">
    ...
</image>
```

there shouldn't be any suffix configured in the image. I did a short build test and that has worked for me

davidcassany:
@schaefi, without it the grub title in the menu of a VMX image some displayname (`displayname="mycustomname"`) configured ended with a grub meny entry like this:

```
mycustomname [VMX]
```

This patch tries to remove the [VMX] part of the menu entry, the patch tries to be a *backport* of that one: https://github.com/SUSE/kiwi/pull/290

davidcassany:
I think the issue only appears when the displayname is long enough to fill the line, so that when adding those extra characters the title string exceeds the line size and first part of the title gets cut.

schaefi:
Ah I remember the fix in the python based kiwi and thought we had this in the perl kiwi too
ok you are right, that's missing here

PR #651, opened 2017-08-29
Return 'string' instead of an object

rpasche:
Commit e75ae3af3588a90db0b6d544fa2f8a7eb866e5e9 changed the returned value
of setupBTRFSSubVolumes to an object resulting in an error to create
the image on btrfs.

### DIFF
--- a/modules/KIWIGlobals.pm
+++ b/modules/KIWIGlobals.pm
@@ -1031,7 +1031,7 @@ sub setupBTRFSSubVolumes {
         }
     }
     if (! %phash) {
-        return $this;
+        return $path;
     }
     $kiwi -> info ("Creating btrfs pool\n");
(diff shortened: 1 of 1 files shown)


rpasche:
To be more precise, the returned "object" results in the following line

https://github.com/openSUSE/kiwi/blob/6748f80e2e0f39cf11dc98ffb04d74e319d26ba9/modules/KIWIImage.pm#L3806

to be evaluated to something like

```
df -h KIWIGlobals (0x123456) 2>&1
```
instead of the "path" of the mounted destination, like

'''
df -h /var/kiwi/destination/path 2>&1
````

schaefi:
yes makes sense and sorry for the very late reply. I didn't notice it

PR #652, opened 2017-09-26
Fix vmdk disk convertion when using LVM

davidcassany:
This patch fixes bsc#1059715. Calling 'pvs --noheadings -o vg_name <device>'
is not sufficient to get the volume name and to determine if LVM is being
used, as it could happen that udev events are not yet processed, thus the
<device> is not yet properly mapped. With this patch blkid is being used to
determine if the type of the requested partition is LVM and if so it proceeds
to a polling strategy to call pvs tool until it succeeds. Timeout is set to
30 seconds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+dist: trusty
+sudo: required
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,3 +19,4 @@ install:
     - cpanm -f -n $(cat .perlmodules | tr "\n" " ")
 
 script: make test
+dist: precise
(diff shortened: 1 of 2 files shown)


davidcassany:
@scheafi, just for the record, the root cause of the issue happens when the raw image is mounted before the conversion to vmdk in order to check the version of vmtoolsd. In order to mount the raw image the raw file is loop mounted and the following call is used in order to make sure udev is up to date and all devices are mapped after 'losteup':

$ udevadm settle --timeout=30 2>&1

This call makes sure all events in the udev event stack are handles before returning, however this is not sufficient, since nothing guarantees that udev events stack is up to date. Udev is event driven, thus new events are also included asynchronously.

davidcassany:
No idea why is it failing... it does not fail locally, trying to figure it out.

PR #656, opened 2017-10-12
- Keep NVMe drivers in the initrd to support systems with the root fi…

rjschwei:
…lesystem

  on NVMe devices


morbidrsa:
I'm not sure you wan't nvmem as well as it's EEPROMs and such, I think nvme only is sufficiant

PR #658, opened 2017-11-09
- Keep Melanox and hyperv kernel modules in the initrd

rjschwei:
+ Due to jitters in boot on Azure the drivers are not always found, keeping
    then in the initrd avoids the issue in the virtualized environment


olafhering:
Actually the bug is (most likely) in kiwi. It just happens to work most of the time:

/init runs `chroot $mnt /preinit'
/preinit runs udevd
udevd does insmod on its own, or calls modprobe, all within $mnt
But mlx4_core does `__request_module`, which in turn also calls /sbin/modprobe
Since the root from kernel point of view is still initramfs that modprobe does not know about $mnt.
As a result mlx4_en can not be found.

PR #659, opened 2017-11-12
KIWIImage: add grub2-mkimage prefix option

ftake:
grub2-mkimage on openSUSE Leap 42.3 now requires "-p" option.
See also ece8cb9e


ftake:
openSUSE bug report: https://bugzilla.opensuse.org/show_bug.cgi?id=1067775

schaefi:
Thanks for the patch it looks good and I'll merge it. But as a side not the last supported target distro for this kiwi version was Leap42.1 and I recommend to use the next generation kiwi http://suse.github.io/kiwi/

PR #660, opened 2017-11-21
no more pam-modules; and prepare for suse_version == 1550

DimStar77:


DimStar77:
Fixed the missing '%' for sle_version variable

PR #661, opened 2017-12-12
Avoid race conditions because of using 'udevadm settle' command

davidcassany:
This commit fixes bsc#1070587. With this patch kiwi tries to populate
the device map when using loop devices for 10 seconds and fails
if after that time no devices are found.

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -2406,7 +2406,14 @@ sub setupBootDisk {
             #==========================================
             # Create loop device mapping table
             #------------------------------------------
-            %deviceMap = $this -> setLoopDeviceMap ($this->{loop});
+            my $it = 0;
+            while (!keys %deviceMap) {
+                if ($it++ >= 5){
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi I added the `needinfo` label since I am awaiting Jozef feedback on bugzilla to verify the patch solves his issue.

I noticed the 'udevadm settle' issue is potentially going to appear in other places also 😞 Not sure if we should try to get rid of all of them, so far it only appears is quite concrete cases.

davidcassany:
Verified that this patch does not solve bsc#1070587, droping the PR, the problem is tightly related to the underlying tools in SLE12 SP2, as the following partitioning only fails in SLE12 SP2 and not in SLE12 SP3;

```bash
qemu-img create image.raw 2048M
losetup --logical-blocksize 4096 -f --show image.raw
fdasd -f /dev/loop0

(p.lxboot: fdasd: n p cur_position +300M w q)

kpartx -s -a /dev/loop0
```

And device mapper does not map the expected partitions. Nothing related with the changed proposed here, closing the PR.

PR #663, opened 2018-02-01
Build initrd for every single kernel installed using dracut

davidcassany:
This commit fixes bsc#1075810.

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -1560,16 +1560,23 @@ function updateModuleDependencies {
     # ---
     local IFS=$IFS_ORIG
     local depmodExec=$(lookup depmod 2>/dev/null)
+    local kernel_version
     if [ ! -x "$depmodExec" ];then
         Echo "Could not find depmod executable"
         Echo "Skipping module dependency update"
         systemIntegrity=unknown
         return
     fi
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi , this is the patch provided by Vladimir Nadvornik in the issue bsc#1075810. The patch looks sane to me and not really invasive. The only thing I wonder if it could ever happen that `uname -r` returns something that is different than any folder present in `/lib/modules`. So not sure if it makes sense to verify `uname -r` value is included in the list form `/lib/modules/*`, probably this is being too paranoid. A part from that the patch looks good to me.

Do you think it makes any sense to port that in to kiwi-ng? I even doubt the related use case is possible with kiwi-ng, so I would not port it.

schaefi:
The patch looks good to me and I think if ```uname -r``` would report a name not matching the directory it would cause many problems with other scripts in the system. But in theory it could be different. Nevertheless I'm ok with that assumption

>Do you think it makes any sense to port that in to kiwi-ng? I even doubt the related use case is >possible with kiwi-ng, so I would not port it.

agreed, the code for that will soon be moved out to its own package and repo and are to my understanding aligned to v7 only.

davidcassany:
@schaefi all right, then I will merge and handle the appropriate maintenance update.

Thanks

PR #664, opened 2018-02-14
Set serial console configuration for grub for ec2 firmware (bsc#1071135)

davidcassany:
This commit sets the serial console for grub2 in case an ec2 firmware
is used. This is a hack to enable serial console for Azure images.

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -319,6 +319,7 @@ sub new {
         $type{type}                   = $xmltype -> getTypeName();
         $type{vga}                    = $xmltype -> getVGA();
         $type{volid}                  = $xmltype -> getVolID();
+        $type{format}                 = $xmltype -> getFormat();
         if (! $type{filesystem}) {
             my $fsro = $xmltype -> getFSReadOnly();
             my $fsrw = $xmltype -> getFSReadWrite();
@@ -4610,7 +4611,7 @@ sub setupBootLoaderConfiguration {
 
         # Set serial console mode for ec2* firmwares. This is a hack for
         # Azure images in order to avoid new attributes in XML schema.
-        if ($type->{firmware} =~ m/ec2/) {
+        if ($type->{firmware} =~ m/ec2/ && $type->{format} eq 'vhd-fixed') {
             $console_mode = 'serial'
         }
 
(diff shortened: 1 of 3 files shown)


modules/KIWIBoot.pm
```diff
@@ -4608,6 +4608,12 @@ sub setupBootLoaderConfiguration {
         # console (console) or the serial terminal (serial)
         my $console_mode = 'graphics';
 
+        # Set serial console mode for ec2* firmwares. This is a hack for
+        # Azure images in order to avoid new attributes in XML schema.
+        if ($type->{firmware} =~ m/ec2/) {
```
rjschwei: Also needs the condition for $type->{format} eq 'vhdfixed'

Then we catch both Azure and EC2

modules/KIWILinuxRC.sh
```diff
@@ -3162,6 +3162,14 @@ EOF
     else
         echo "GRUB_TERMINAL=console"  >> $inst_default_grub
     fi
+
+    # Set serial console mode for ec2* firmwares. This is a hack for
+    # Azure images in order to avoid new attributes in XML schema.
+    if [[ "$kiwi_firmware" =~ "ec2" ]]; then
```
rjschwei: Also needs the format check for vhdfixed.

davidcassany: Here we do not have any $kiwi_format variable available, checking how it can be included here is needed for this change.

davidcassany:
@rjschwei would it be possible for you test it? I am having issues to boot the image with qemu and so far I could just check the disk contains the appropriate grub2.cfg and that the .profile file includes the format for the boot code.

I have no experience with ec2 and I have no idea how should I test it in my local env.

davidcassany:
I have added block label to make sure it is not merged. The patch is not sufficient, grub is not configured properly after the first boot. The boot code sets the /etc/default/grub file correclty so I can see the GRUB_SERIAL_COMMAND variable with the correct values; however after running the grub2-mkconfig (it is executed by the boot code on a later stage) the resulting file does not contain this information.

davidcassany:
Right, finally found the that GRUB_SERIAL_COMMAND is only applied when GRUB_TERMINAL is set to serial. Now I have successfully tested in ec2 firmware and with vhd-fixed format.

davidcassany:
@rjschwei, @schaefi when you approve the PR I will proceed with the maintenance update. 

Thanks

schaefi:
Looks great 👍

PR #665, opened 2018-03-01
Fixup elog shell pid detection

schaefi:
Startup of the debug shell happened as extra process from
a setctsid call. The lookup of the process was based on
a fuser call on the slelected tty. However it can happen
that another process is also connected to the tty which
does not allow for a stable detection of the debug shell
process. This patch runs the debug shell as sub process
of the calling terminal and uses the sub process pid
This Fixes bsc#1075813


nadvornik:
LGTM

PR #667, opened 2018-09-25
GCE name

rjschwei:
Fix the name of the tar archive generated for GCE images, they should not all be named the same.


schaefi:
@rjschwei The patch looks good to me, before I merge one short question. Have you built an image with this change applied ? We have no integration test build matrix for the legacy kiwi, that's why I ask. Thanks

rjschwei:
Yes, I built an image, location in private e-mail as that is in IBS.

PR #669, opened 2018-10-03
integrate last translation changes

stanislav-brabec:
Kiwi is no more maintained and it is being decommissioned from the Weblate.

Please integrate remaining translation changes that were done between disabling push access to opensuse-i18n and switching the translation project read-only.


schaefi:
Thanks much. Looks good to me :+1:

PR #670, opened 2018-11-19
Replace 'ifplugd' with 'auto' in STARTMODE

Bischoff:
This PR replaces
```
STARTMODE='ifplugd'
```
with
```
STARTMODE='auto'
```
in `/etc/sysconfig/network/ifcfg-ethN` as prepared by Kiwi.

Reason is that `ifplugd` seems to be deprecated on SLE family (at least I could not find it in SLE 12 SP3, while it still exists on Leap 15.0 ), probably replaced by `wicked`. It was already removed from `template/ix86/suse-SLE12-JeOS/config.xml` in commit https://github.com/openSUSE/kiwi/commit/c1c1c602245348f725d41d1457de85aa5a6f3f34 from @schaefi . On openSUSE, `auto`should just work fine, and is apparently the default anyway.

If the SLE machine prepared by Kiwi contains the wrong setting, then it is not completely properly functional after a reboot. For example, it has no link-local IPv6 address.

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -3452,7 +3452,7 @@ function setupDefaultPXENetwork {
     mkdir -p $prefix/etc/sysconfig/network
     cat > $niface < /dev/null
     echo "BOOTPROTO='dhcp'"    >> $niface
-    echo "STARTMODE='ifplugd'" >> $niface
+    echo "STARTMODE='auto'"    >> $niface
     echo "USERCONTROL='no'"    >> $niface
 }
 #======================================


schaefi:
The reason for ifplugd was this bug years ago:

* https://bugzilla.suse.com/show_bug.cgi?id=365170

I'd say the world changed a lot since then and I'm ok with the changes proposed here

Regarding SLE15 we need to change the repo from this legacy area to SUSE/kiwi-descriptions
repo and probably apply the changes suggested here in that space too: See

custom_boot/functions.sh

which also still uses ifplugd for the kiwi netboot stuff.

However customers using the dracut based network deployment will benefit from the network
code in dracut

davidcassany:
Great then all looks good.

@Bischoff just one more thing, I guess we need to make a MR for that as you need it for SLE12-SP3. Would you mind opening a bugzilla ticket if that's the case? So that we can refer to a ticket in changelog.

Bischoff:
Sure @davidcassany . Doing it now.

PR #672, opened 2019-02-10
Sort filesystem listings

bmwiedemann:
Sort filesystem listings
so that kiwi works in a reproducible way
in spite of indeterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

### DIFF
--- a/modules/KIWICollect.pm
+++ b/modules/KIWICollect.pm
@@ -2091,7 +2091,7 @@ sub collectProducts {
             $this->logMsg('I', "No products found, skipping");
             next RELEASEPACK;
         }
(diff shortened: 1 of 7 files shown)


davidcassany:
Former kiwi is in maintenance mode and only alive in SLE12 series and in Leap 42.3. Unless it is known to cause some specific issues, in general, I would be more in favor of not touching this code. The change does not look too intrusive to me so we could probably merge it here, but I doubt if this deserves a specific backport to SLE.

@schaefi any thoughts regarding that?

schaefi:
I agree, we can merge this change, Thanks Bernhard for the work but we won't be porting this to SLE12 unless there is a customer L3 report available.

@davidcassany feel free to merge

PR #674, opened 2019-05-08
Set SERVER variable as expected by netboot code

davidcassany:
This commit sets $host from $SERVER variable if present and no
$kiwiserver is given from the command line.

The same criteria applies for $type and $SERVERTYPE variables.

Fixes bsc#1133537


davidcassany:
@schaefi I added the blocked label as I'd like to validate it with the issue reporter first. However I believe this is ready for a review.

schaefi:
yes we should wait for positive feedback before merge. I'm fine with that

schaefi:
feel free to merge when you see fit. Thanks

PR #675, opened 2019-05-31
Normalize partedGetPartitionID output (bsc#1136744)

davidcassany:
This commit ensures the partedGetPartitionID output returns the
same partition ID codes with or without the presence of sgdisk
utility. It basically returns the sgdisk parsed value in lowercase.

Additionally this commit also makes sure that the partition stated
to be of `ef` type in `config.<mac>` is named `legacy`, to be consistent
with the partedGetPartitionID expectations.

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -6026,10 +6026,9 @@ function fdasdGetPartitionID {
 function partedGetPartitionID {
     # /.../
     # prints the partition ID for the given device and number
-    # In case of a GPT map to the GUID code from the sgdisk
-    # utility. If sgdisk is not available map to the kiwi
-    # fdisk compatible hex id's which uses ee for any kind
-    # of unknown GPT partition entry
+    # In case of GPT it maps ID values according to the name kiwi
+    # defines for the partition. In case of unknown name try to use
+    # sgdisk, if available, or just default to `ee`
     # ----
     local IFS=$IFS_ORIG
     local parted=$(parted -m -s $1 print | grep -v Warning:)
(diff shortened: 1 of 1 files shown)


schaefi:
The coding looks good to me, but I think only a real live test can make sure nothing else breaks. I guess you are waiting for feedback from Vladimir.

Thanks much for you effort here

davidcassany:
@schaefi yes that's why this PR is blocked with needinfo. Also I did not ask your review because of that, at first I want some real tests. I'll try reproduce it also, Vladimir said we will test it on Monday and gave me some hints to reproduce it. Vladimir has a different suggestions to fix it, but I am afraid to touch `partedGetParitionID` method return values for the default GPT case, I have no clue what else could break in OEM realm.

modules/KIWILinuxRC.sh
```diff
@@ -10493,6 +10494,10 @@ function pxePartitionInputGeneric {
             partID=fd
             pname=lxroot
         fi
+        if [ $partID = "ef" ] && [ ! -z "$RAID" ];then
```
nadvornik: My test fails on this: [ ! -z "$RAID" ]

davidcassany: Damn sorry, just realized I did not push this change last Friday, this was supposed to be 

```bash
if [ $partID = "ef" ] && [ "$partMount" = "/boot/efi" ];then
```

davidcassany:
@nadvornik with the current status the PXE image just worked fine with and without sgdisk inside the initrd. Could you test it again?

The solution I did is to trust partition names over the sgdisk output. I noted that the code does some assumptions about parted default values and it's GPT support that I believe are not valid anymore. For instance the code assumes that the default partition type is `83` however in my tests it is `07`, it also assumes that for GPT partitions parted can't set the partition type, thus it never sets the partition type partitions.

Because of that I believe the best approach to check the partition type is relaying on the partition name, which is completely set by KIWI (depending on its type) and tool independent. Trying to run parted in a consistent way with sgdisk and GPT partitions I believe would be error prone and lead to a relevant refactoring of the related code. Since this is a very sensitive portion of code, that can only be validated running the initrd, huge changes there should be avoided when possible.

nadvornik:
LGTM

PR #676, opened 2019-07-09
Fix handling of zypper return codes

schaefi:
Only the following codes should be handled as errors:
104 - ZYPPER_EXIT_INF_CAP_NOT_FOUND
105 - ZYPPER_EXIT_ON_SIGNAL
106 - ZYPPER_EXIT_INF_REPOS_SKIPPED

### DIFF
--- a/modules/KIWIManager.pm
+++ b/modules/KIWIManager.pm
@@ -296,6 +296,18 @@ sub setupScreenCall {
     #==========================================
     # check exit code from session
     #------------------------------------------
+    if (($code != 0) && ($this->{manager} eq 'zypper')) {
+        if (($code == 104) || ($code == 105) || ($code == 106)) {
+            # Treat the following exit codes as error
+            # 104 - ZYPPER_EXIT_INF_CAP_NOT_FOUND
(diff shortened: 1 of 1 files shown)


modules/KIWIManager.pm
```diff
@@ -296,6 +296,18 @@ sub setupScreenCall {
...
     #------------------------------------------
+    if (($code != 0) && ($this->{manager} eq 'zypper')) {
+        if (($code == 104) || ($code == 105) || ($code == 106)) {
+            # Treat the following exit codes as error
+            # 104 - ZYPPER_EXIT_INF_CAP_NOT_FOUND
+            # 105 - ZYPPER_EXIT_ON_SIGNAL
+            # 106 - ZYPPER_EXIT_INF_REPOS_SKIPPED
+            $code = 1;
+        } else {
+            # Treat all other 100 codes as non error codes
```
davidcassany: Shouldn't we consider any error code under 100 as a regular error?

schaefi: yes my bad, should be better now

PR #677, opened 2019-07-16
Improve partition validation for PXE

davidcassany:
This commit includes a patch to relax the partition
size validation. It turned to happen that under some
specific partition size configurations in the PXE client
the partition layout was always seen as an invalid one
in each boot. This was causing a repartition and redeploy
of the PXE image in every single boot.

Fixes bsc#1141156


davidcassany:
@schaefi please see the full story in https://bugzilla.opensuse.org/show_bug.cgi?id=1141156

schaefi:
Looks good to me, but I think we need the same patch also in kiwi-descriptions/custom_boot/functions.sh

davidcassany:
> Looks good to me, but I think we need the same patch also in kiwi-descriptions/custom_boot/functions.sh

Yes I am having  a look to that.

PR #681, opened 2019-11-13
fix sha256 checksum file creation

seife:
This got broken in commit 166cd36d0c79 ("Sort filesystem listings...")
Also fixes "Useless use of sort in scalar context..." runtime warning.


schaefi:
Thanks makes perfect sense to me :+1:

schaefi:
I'll merge and provide new package on Virt:App project

PR #682, opened 2021-02-01
Update shim.efi file locations

davidcassany:
This commit updates the paths from which the shim efi files are
extracted from the install initrd in order to build the ISO installation
media. Since shim v15 '/usr/lib64/efi/shim.efi' is a symlink pointing to
'/usr/share/efi/shim.efi', hence '/usr/share/efi' folder should be
extracted too.

Fixes bsc#1180067

### DIFF
--- a/modules/KIWIBoot.pm
+++ b/modules/KIWIBoot.pm
@@ -3629,6 +3629,7 @@ sub setupBootLoaderStages {
         }
         if ($firmware eq "uefi") {
             $stages{efi}{data}      = "'usr/$lib/efi/*'";
+            $stages{efi}{data_share}= "'usr/share/efi/*'";
             $stages{efi}{shim_ms}   = "usr/$lib/efi/shim.efi";
             $stages{efi}{shim_suse} = "usr/$lib/efi/shim-opensuse.efi";
             $stages{efi}{signed}    = "usr/$lib/efi/grub.efi";
@@ -3939,6 +3940,7 @@ sub setupBootLoaderStages {
         if (($type eq 'iso') && ($firmware eq "uefi")) {
             $kiwi -> info ("Importing grub2 shim/signed efi modules");
             my $s_data      = $stages{efi}{data};
+            my $s_data_share= $stages{efi}{data_share};
             my $s_shim_ms   = $stages{efi}{shim_ms};
             my $s_shim_suse = $stages{efi}{shim_suse};
             my $s_signed    = $stages{efi}{signed};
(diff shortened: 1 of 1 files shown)


davidcassany:
There is a test build here https://build.suse.de/package/show/home:dcassany:SLE-12-SP3:Update/kiwi-templates-SLES12-JeOS.11574

I tested in my local env builds of SLE-12-SP3 using shim v14 and v15 and the build was successful for both cases.

schaefi (approved):
Great you rock :+1: Really much appreciated

schaefi:
I will merge this now and take care for a package submission if you don't have any objections ? I think working on this old code base was not so much fun so it would be fair if you don't have to care for the packaging stuff :) ok ?

davidcassany:
> I will merge this now and take care for a package submission if you don't have any objections ? I think working on this old code base was not so much fun so it would be fair if you don't have to care for the packaging stuff :) ok ?

Fine with me, thanks. But I would have done it without any objection or bad feeling :wink: 

Anyway I noticed there is no kiwi in SP3, hence I assume we should submit it to SP2 and SP4. This way SP3 and higher will includw the fix. Also I noticed that the branches of this repo are slightly deviated from what is released.

schaefi:
> Fine with me, thanks. But I would have done it without any objection or bad feeling 

I know and this is great. I just felt a little bit bad that I made you digging into this old perl code ;) Also I've learned how to change the author e-mail in git commits. This one had david@localhost :)

I've created the two submissions to SLE12-SP2 and SP4 (since SP3 is the same as SP2 this will also fix it for SP3), SP4 is the same as SP5 so it gets fixed there too.

I'll update the bug with the info

Thanks much

schaefi:
I've also updated Virt:App with the latest legacy kiwi (don't think it will be used though)

davidcassany:
There is a test build here https://build.suse.de/package/show/home:dcassany:SLE-12-SP3:Update/kiwi-templates-SLES12-JeOS.11574

I tested in my local env builds of SLE-12-SP3 using shim v14 and v15 and the build was successful for both cases.

> > Fine with me, thanks. But I would have done it without any objection or bad feeling
> 
> I know and this is great. I just felt a little bit bad that I made you digging into this old perl code ;) Also I've learned how to change the author e-mail in git commits. This one had david@localhost :)
> 
> I've created the two submissions to SLE12-SP2 and SP4 (since SP3 is the same as SP2 this will also fix it for SP3), SP4 is the same as SP5 so it gets fixed there too.
> 
> I'll update the bug with the info
> 
> Thanks much

ouch I guess this is because I have not been using this repo for quite a while... Some months ago I wiped my global git user name to force myself to configure it on repo basis, I guess I forgot to update this repository.

PR #684, opened 2022-10-03
Try up to 5 retries of wicked dhcp query

aaannz:
In some devices it appears waiting for link and couple more seconds is not enough and wicked dhcp call fails. tcpdump showed that DHCP request is not sent, so retry whole dhcp query up to 5 times in order to get the IP address.

### DIFF
--- a/modules/KIWILinuxRC.sh
+++ b/modules/KIWILinuxRC.sh
@@ -5162,11 +5162,16 @@ function setupNetworkWicked {
         # try DHCP_DISCOVER on all interfaces
         if setIPLinkUp $try_iface; then
             dhcp_info=/var/run/wicked/wicked-${try_iface}.info
-            $wicked_dhcp4 --debug all \
-                --test --test-output $dhcp_info $try_iface
-            if [ $? = 0 ] && [ -s $dhcp_info ];then
-                DHCPCD_STARTED="$DHCPCD_STARTED $try_iface"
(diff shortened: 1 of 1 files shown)


aaannz:
CC @davidcassany

schaefi (approved):
Looks good to me, thanks

seq belongs to coreutils which is part of the kiwi initrd description files. So this is safe

I like @davidcassany to approve as well but I think this is good to go

schaefi:
One aspect I noted is this: ```/usr/lib/wicked/bin/wickedd-dhcp4 --help```

```
   --test-timeout <timeout in sec> (default: 20+10)
```

Maybe the code can be done simpler by using the timeout option ?

aaannz:
> One aspect I noted is this: `/usr/lib/wicked/bin/wickedd-dhcp4 --help`
> 
> ```
>    --test-timeout <timeout in sec> (default: 20+10)
> ```
> 
> Maybe the code can be done simpler by using the timeout option ?

I did not check the internals of wicked, but my understanding of this is that it is time wicked waits for the response. However during debugging I asked user for packet capture on their dhcp server and no dhcp requests even arrived (excluding PXE client queries). So we need wicked to send new dhcp requests.

schaefi:
makes sense, thanks for the explanation :+1:

davidcassany (approved):

